PR #2, opened 2014-10-02
CMR4926

rhc54:
SVN r32817
Correct the classpath to correctly include the local directory so Java programs find the application class

SVN r32823
Cover the remaining code paths for Java apps to define class path


rhc54:
@osvegis  - can you please review?

osvegis:
reviewed
It's ok. I checked it with and without -cp parameter and CLASSPATH variable.

rhc54:
RM-approved

PR #3, opened 2014-10-02
fortran: fix the MPI_SIZEOF problems in the Fortran interfaces

jsquyres:
This PR is the union of several commits on master; I squashed them down to a single commit here, for simplicity.

Refs open-mpi/ompi#198


jsquyres:
Craig reviewed this patchset as ok, but now github says it can't merge, so I think I need to rebase this to the top of the v1.8 branch...

jsquyres:
Ok, rebased these two commits; github is happy again.  This PR is good-to-go.

PR #10, opened 2014-10-06
OSHMEM: fix memheap/sshmem/mmap to use MAP_PRIVATE instead of SHARED to speedup registration

mike-dubman:


jsquyres:
@miked-mellanox tells me in IM that this has been reviewed.

PR #11, opened 2014-10-07
oob/tcp: fix a race condition

ggouaillardet:
Mimick the btl/tcp protocol to solve the race condition that happens
when two peers try to connect to each other at the same time

cmr=v1.8.4:reviewer=rhc

This commit was SVN r32799.
(cherry picked from commit 9661e4537fc6d932c63e4680b23cfd86a3f64035)


rhc54:
No real harm in bringing this in, though the situation it fixes should never occur in the 1.8 series.

PR #13, opened 2014-10-07
osc/rdma uses counters to determine if all messages have been received

hjelmn:
before exiting synchronization calls. The problem is that the active
target counter is always increasing (never zeroed). If over 2^31-1
messages are sent this causes the counter to overflow (in itself this
isn't an error). This causes test/wait to return before the communication
is complete. There is an additional error in the use of the fragment
flush function. If PSCW synchronization is in use this function CAN NOT
be called unless a post message has arrived.

Relevant mailing list thread: http://www.open-mpi.org/community/lists/devel/2014/10/16016.php

This commit fixes both issues. Tested against MTT and issue reproducer.


jsquyres:
@hjelmn Who did you want to review this PR?

hjelmn:
Master hash open-mpi/ompi@eed7b45db59b356f6257f3b019e77643f14f84ce

hjelmn:
@jsquyres Not many besides myself know the OSC code. Either you or dave could probably review it. I added the reproducer in ompi-tests. c_post_start

jsquyres:
Nathan: use the "open-mpi/ompi@[hash]" syntax to get appropriate cross-linking in the Github web UI.

Thanks!

goodell:
I took a quick look at this.  I didn't see any obvious coding mistakes or anything like that, but I don't really have time to grok the one-sided impl again to give this a really thorough review.

goodell:
Actually, I did spot one issue that isn't in this commit/PR directly.  The `active_incoming_frag_count` and similar fields are all `int32_t` types that are intended to overflow.  But signed overflow in C is undefined.  I think you should either manually implement the overflow logic or convert these types to unsigned.

hjelmn:
Was after a minimal change but I see no problem with also addressing Dave's comments. I will fix in master then refile this CMR.

hjelmn:
Closing this so I can refactor my branch and refile.

PR #15, opened 2014-10-08
memchecker: per RFC, use calloc for OBJ_NEW

jsquyres:
I think that this is worthy of bringing to v1.8 because it will significantly reduce valgrind output for those who do `--enable-memchecker` builds.  Hence, it can easily be considered a bug fix, not an enhancement.

---

With --enable-memchecker builds, use calloc(3) for OBJ_NEW instead of malloc(3).  This cuts down on a lot of valgrind/memory checker false positive output.

Also make a minor change in the valgrind configure.m4; have it assign 0xf to a char.  The prior assignment (of 0xff) was warning about an overflow.  This didn't really matter, but we might as well make the test not have a gratuitious warning in it.

(cherry picked from commit a422d893b8e63347deb71bbaaa14ca25ddb1f192)

### DIFF
--- a/opal/class/opal_object.h
+++ b/opal/class/opal_object.h
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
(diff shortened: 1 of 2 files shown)


jsquyres:
@rhc54 Can you review?  Pretty trivial.

PR #21, opened 2014-10-13
Update PMI detection logic

rhc54:
Update the PMI detection logic to support oddball case where both lib and lib64 exist, sometimes seen in Slurm installations. Also, better support Cray installations of all ages

Master commits:

https://github.com/open-mpi/ompi/commit/169a1866b8b3035cd118367077bd9ff620dc7e1e
https://github.com/open-mpi/ompi/commit/b1a58726ac17e596f25d732c97349d6156a9fdd1


rhc54:
@jsquyres Please review

config/opal_check_pmi.m4
```diff
@@ -45,78 +49,85 @@ AC_DEFUN([OPAL_CHECK_PMI],[
...
+           # check for pmi-1 lib */
+           files=`ls -U $opal_check_pmi_install_dir/lib64/libpmi.* 2> /dev/null | wc -l`
+           AS_IF([test "$files" -gt "0"],
+                 [opal_have_pmi1=1
+                  AS_IF([test "$opal_default_loc" = "no"],
+                        [opal_check_pmi_$1_LDFLAGS="-L$opal_check_pmi_install_dir/lib64"
+                         opal_pmi_rpath="$opal_check_pmi_install_dir/lib64"
+                         opal_pmi_added_ldflag=yes])
+                  opal_check_pmi_$1_LIBS="-lpmi"],
+                 [files=`ls -U $opal_check_pmi_install_dir/lib/libpmi.* 2> /dev/null | wc -l`
```
jsquyres, line 74: Is the -U necessary here?  It seems to have different meanings between Linux and OS X.

@jsquyres It is purely there for speed and can be removed

config/opal_check_pmi.m4
```diff
@@ -45,78 +49,85 @@ AC_DEFUN([OPAL_CHECK_PMI],[
...
+                  AS_IF([test "$files" -gt "0"],
+                        [opal_have_pmi1=1
+                         AS_IF([test "$opal_default_loc" = "no"],
+                               [opal_check_pmi_$1_LDFLAGS="-L$opal_check_pmi_install_dir/lib"
+                                opal_pmi_rpath="$opal_check_pmi_install_dir/lib"
+                                opal_pmi_added_ldflag=yes])
+                         opal_check_pmi_$1_LIBS="-lpmi"])])
+           # check for pmi.h
```
jsquyres, line 82: After you find the correct directory, shouldn't we AC_CHECK_LIB to ensure that the library is valid?  It looks like the AC_CHECK_LIB code was deleted.

@jsquyres I viewed that as being an unnecessary complication - the code is already getting rather hard to follow, and checking the lib doesn't seem worth the trouble.

config/opal_check_pmi.m4
```diff
@@ -45,78 +49,85 @@ AC_DEFUN([OPAL_CHECK_PMI],[
...
+           AS_IF([test -f "$opal_check_pmi_install_dir/include/pmi.h"],
+               [AS_IF([test "$opal_default_loc" = "no"],
+                      [opal_check_pmi_$1_CPPFLAGS="-I$opal_check_pmi_install_dir/include"
+                       opal_pmi_added_cppflag=yes])],
+               # this could be SLURM, which puts things in a different location
+               [AS_IF([test -f "$opal_check_pmi_install_dir/include/slurm/pmi.h"],
+                       # even if this was the default loc, we still need to add it in
+                       # because of the slurm path addition
+                      [opal_check_pmi_$1_CPPFLAGS="-I$opal_check_pmi_install_dir/include/slurm"
+                       opal_pmi_added_cppflag=yes])])
```
jsquyres, line 92: Shouldn't we AC_CHECK_HEADER to ensure that pmi.h is valid, too?

@jsquyres As above, doesn't seem worth the trouble.

jsquyres:
Just for the github thread: this PR was replaced by #53.

PR #27, opened 2014-10-14
contrib: added the check-help-strings.pl file.

alinask:
(from master - commits: 98ef295c, 7a3adcf0, c59ca9c7, 82b49c2b)

@miked-mellanox @jladd-mlnx

### DIFF
--- a/contrib/check-help-strings.pl
+++ b/contrib/check-help-strings.pl
@@ -1,0 +1,322 @@
+#!/usr/bin/env perl
+#
+# Copyright (c) 2014 Cisco Systems, Inc.  All rights reserved.
+# $COPYRIGHT$
+#
+# Simple script to check all the opal_show_help (and orte_show_help)
+# strings against what is found in help files.
+#
+
+use strict;
+
+use Cwd;
+use File::Find;
+use Getopt::Long;
+use Data::Dumper;
+
+my $num_warnings = 0;
+my $num_errors = 0;
+
+###########################################################################
+
+my $VERBOSE = 0;
+my $HELP = 0;
+
+GetOptions(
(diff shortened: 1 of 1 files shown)


rhc54:
I fail to see a reason why this file needs to go into the production release tarball - we have it in the developer's trunk, which would seem adequate. Am I missing something? It certainly isn't a "bug"!

mike-dubman:
@rhc54 ,
- we run this script as part of the jenkins routine for every commit/pr pushed to the any branch.
- w/o this fix, scripts fails and failing the jenkins.
- We can add extra logic to the jenkins script to bypass branches with old script interface but it does not seem right and complicates jenkins logic.

@jsquyres, @rhc54 - maybe all "support" scripts should be pushed to separate repo and it will be less hassle to support branches, no need to sync scripts from master to vX.Y on every script change?

Thanks

jsquyres:
@rhc54 FWIW, I don't see a problem including this script on the v1.8 branch.

@miked-mellanox I'm still not convinced that a separate script repo is a good idea.  In some cases, there are definitely differences between the different release series for the various scripts.  I think there's pluses and minuses for both options (i.e., having the scripts in each branch and having one central set of scripts).  Meaning: I don't think there's necessarily more plusses to having them in a single/central repo.

rhc54:
Here's another thought - I would have accepted it if someone had simply provided even a tiny bit of rationale behind the PR instead of just "add this because I said so". This is a _stable_ release branch, so there has to be _some_ justification for adding something - if you can't provide a reasonable explanation, then you should expect it to be rejected.

mike-dubman:
@rhc54 - sorry about that, next PR will be better equipped.

PR #33, opened 2014-10-15
var: fix segv in deprecated file var show_help()

jsquyres:
Ensure to include the new variable filename in the show_help() output
when we load a deprecated MCA param from a file.

Fixed #236

(cherry picked from commit dc66e197cc48396002d0dcf6f5f04dc6b581d7fd)


jsquyres:
Reviewed by @hjelmn

PR #34, opened 2014-10-18
Fix building system: correct libltdl check for advise capability.

artpol84:
ompi commit: open-mpi/ompi@3ff16e7da7610a68c50ec89ef986e17191f03f3b
ompi pull request: https://github.com/open-mpi/ompi/pull/240

### DIFF
--- a/config/opal_setup_libltdl.m4
+++ b/config/opal_setup_libltdl.m4
@@ -170,7 +170,7 @@ AC_DEFUN([_OPAL_SETUP_LIBLTDL_INTERNAL],[
         OPAL_LIBLTDL_INTERNAL=1
 
         CPPFLAGS_save="$CPPFLAGS"
-        CPPFLAGS="-I."
+        CPPFLAGS="-I$srcdir/opal/libltdl/"
         # Must specifically mention $srcdir here for VPATH builds
         # (this file is in the src tree).
         AC_EGREP_HEADER([lt_dladvise_init], [$srcdir/opal/libltdl/ltdl.h],


jsquyres:
Artem --

I'm struggling to remember why the initial `-I.` was there.  In my comment on your master PR (https://github.com/open-mpi/ompi/pull/240#discussion_r19020450), I was only referring to your new second `-I`, not the first/original `-I.`.

Can you check to see exactly which `gcc -E ...` command is emitted, and see if changing `-I.` to `-I$srcdir` is necessary?

FWIW: the libltdl.h file should be in the source tree (vs. the build tree), so `-I$srcdir` might actually be correct.  But I'm not remembering exactly what that `-I.` was for -- i.e., was it actually for libltdl.h, or was it for something else?  And if it was actually for libltdl.h, is it just plain wrong, and your now `-I$srcdir/...` is the correct one?  I.e., should the original `-I.` (now `-I$srcdir`) just be deleted?

artpol84:
That is a good point, Jeff.

Looking closely to the configuration system I can see that we deal with following conftest.c template:

``` bash
CPPFLAGS="-I$srcdir/ -I$srcdir/opal/libltdl/"
# Must specifically mention $srcdir here for VPATH builds
# (this file is in the src tree).
cat confdefs.h - <<_ACEOF >conftest.$ac_ext
/* end confdefs.h.  */
#include <$srcdir/opal/libltdl/ltdl.h>

_ACEOF
```

That will turn into

``` c
//... < lots of defines > ....
#define OMPI_MPI_CONTRIBS "libompitrace, vt"
#define OPAL_C_HAVE_VISIBILITY 1
#define HAVE_DLFCN_H 1
#define LT_OBJDIR ".libs/"
/* end confdefs.h.  */

#include < ./opal/libltdl/ltdl.h > 
```

For local config and to 

``` c
...
#include < /some/vpath/path/opal/libltdl/ltdl.h >
```

for VPATH.

So we don't actually need "-I." or "-I$srcdir" at all because conftest.c always contains correct path relative to the current configuration directory. 
I updated my pull branch by changing the commit of interest.

artpol84:
I guess I need to go back to ompi and apply this patch there? Additionall pull request?

jsquyres:
Wait, let's figure this out before doing anything...

If we don't need the `-I`, then why did adding it fix your problem?

artpol84:
Sure,
I didn't add -I. I proposed to add -I$srcdir/opal/libltdl/. And now I think that we can also remove "-I."

jsquyres:
I know you didn't add `-I.` -- I just said `-I`, meaning the `-I$srcdir...` one.  :-)

I was responding to your prior comment:

> So we don't actually need "-I." or "-I$srcdir" at all ...

So if we need _neither_ `-I`, then why did your adding `-I$srcdir...` fix the problem for you?

artpol84:
Because in conftest.c we include `ltdl.h` using its full path. And  `ltdl.h` includes few others:

``` c
#include <libltdl/lt_system.h>
#include <libltdl/lt_error.h>
```

They are located in `opal/libltdl/libltdl` and are reffered by relative paths and inaccessible for gcc without corresponding `-I$srcdir/opal/libltdl/`.

P.S. Check the latest patch to clarify my current version.

jsquyres:
Ok, got it -- we _do_ need the `-I$srcdir...`.  Got it.

Yes, please push the final change back to master (i.e., delete the `-I.`), put the master hash here (per the requirement that we have to have all the relevant commits/functionality on master before bringing to v1.8, and then we'll be good to go.

artpol84:
Done: open-mpi/ompi@ce7102cb54360a3ecc47d3d23a972b3c56621c23

PR #44, opened 2014-10-22
Minor usnic fixes

jsquyres:
Two commits:
1. Do not send zero-length modex messages if there are no usnic devices.  On master, open-mpi/ompi@ec4268b59cf42749b3975acc75be013e31925cf8
2. Do not error if a peer did not supply a usnic modex message -- it just means that there are no usNIC BTL modules in that peer.  This has actually be on master/trunk for forever; I just noticed today that it was not on the v1.8 branch.

Reviewed by @goodell

### DIFF
--- a/ompi/mca/btl/usnic/btl_usnic_component.c
+++ b/ompi/mca/btl/usnic/btl_usnic_component.c
@@ -240,32 +240,33 @@ static int usnic_modex_send(void)
     size_t size;
     ompi_btl_usnic_addr_t* addrs = NULL;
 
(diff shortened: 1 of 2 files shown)


jsquyres:
This is assigned to 1.8.4 because, per discussion with @rhc54 this morning, it's a bit of a scalability issue: the fact that the usnic BTL is adding zero-length usnic modex messages is adding to to the overall modex message.  Therefore, fixing it is a candidate for 1.8.4.

PR #45, opened 2014-10-22
btl/sm: re-enable the use of CMA and knem

hjelmn:
At some point we added a sanity check to the btl base to ensure that
the btl flags match the available functions (this prevents user's from
specifying get or put when no function exists). This check was
disabling get for the sm btl since at the time of the check there is
no btl_get function. The simplest fix is to set a dummy value to btl_get
that will be overwritten with the proper value on btl initialization.

Includes master references:
open-mpi/ompi@d7c7bb399388e7f85aa856c38865c8062bf7c0e6
open-mpi/ompi@998e69a6facdecdaa95996ac996b5a249c42f8e0

### DIFF
--- a/ompi/mca/btl/sm/btl_sm_component.c
+++ b/ompi/mca/btl/sm/btl_sm_component.c
@@ -142,13 +142,23 @@ static inline unsigned int mca_btl_sm_param_register_uint(
     return *storage;
 }
 
+#if OPAL_BTL_SM_HAVE_KNEM || OPAL_BTL_SM_HAVE_CMA
+static void mca_btl_sm_dummy_get (void)
+{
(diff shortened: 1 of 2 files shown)


hjelmn:
Reference open-mpi/ompi#239

hjelmn:
@jsquyres Can you please review?

jsquyres:
I just put a comment about the code here: https://github.com/open-mpi/ompi/commit/d7c7bb399388e7f85aa856c38865c8062bf7c0e6#commitcomment-8265592

hjelmn:
@abouteiller I brought over your fixes with the enhancements suggested by Jeff.

PR #47, opened 2014-10-22
osc/rdma: bring over one-sided fixes from master

hjelmn:
This commit brings over two commits:

open-mpi/ompi@eed7b45db59b356f6257f3b019e77643f14f84ce

osc/rdma: fix issue identified by Berk Hess

osc/rdma uses counters to determine if all messages have been received
before exiting synchronization calls. The problem is that the active
target counter is always increasing (never zeroed). If over 2^31-1
messages are sent this causes the counter to overflow (in itself this
isn't an error). This causes test/wait to return before the communication
is complete. There is an additional error in the use of the fragment
flush function. If PSCW synchronization is in use this function CAN NOT
be called unless a post message has arrived.

Relevant mailing list thread: http://www.open-mpi.org/community/lists/devel/2014/10/16016.php

This commit fixes both issues. Tested against MTT and issue reproducer.

open-mpi/ompi@23dd3af946dbfec55f3261c2a1a8e3ea30ed695c

osc/rdma: use unsigned types for all counters

Some of the counters used by the "rdma" one-sided component are intended
to overflow. Since overflow behavior is undefined for signed integers in
C it is safer to use unsigned integers here.

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma.h
+++ b/ompi/mca/osc/rdma/osc_rdma.h
@@ -423,8 +423,8 @@ static inline void mark_incoming_completion (ompi_osc_rdma_module_t *module, int
 {
     if (MPI_PROC_NULL == source) {
         OPAL_OUTPUT_VERBOSE((50, ompi_osc_base_framework.framework_output,
-                             "mark_incoming_completion marking active incoming complete. count = %d",
-                             (int) module->active_incoming_frag_count + 1));
+                             "mark_incoming_completion marking active incoming complete. count = %d. signal = %d",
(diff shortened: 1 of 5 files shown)


hjelmn:
This replaces pull request #13. In addition to the original commit this addresses the issue identified by @goodell

jsquyres:
Does it need a new review from @goodell?

goodell:
Looks OK to me, though I would prefer that if you are bringing over two commits, just bring over two commits.  There's no reason to squash them into a single commit now that we have Git.  That's the role of the merge commit that will be created when Ralph approves the PR.

hjelmn:
@goodell Redid the branch with cherry-picks

goodell:
Looks OK to me.

PR #48, opened 2014-10-23
backport OPAL_WANT_MEMCHECKER macro

ggouaillardet:
This fixes commit open-mpi/ompi-release@3156f98d358a5d3f34e21324a11708441153268f
which is a cherry pick from open-mpi/ompi@a422d893b8e63347deb71bbaaa14ca25ddb1f192

@jsquyres could you please review this PR ?

PR #49, opened 2014-10-23
Fix an other typo introduced by transfer of trunk change to v1.8

ggouaillardet:
and not fixed by open-mpi/ompi-release@c7e9cf0a3cf261974ef082c07f02330112e87f01

@rhc54 could you please review this PR ?


jsquyres:
Same as #48 -- good catch / WTF didn't my compiler catch this? :-(

PR #53, opened 2014-10-24
Simplify PMI configure logic. Master commit:

rhc54:
open-mpi/ompi@75d8a7f

@jsquyres please review


config/opal_check_pmi.m4
```diff
@@ -21,128 +21,217 @@
...
+# and validity of a PMI library
+#
+# OPAL_CHECK_PMI_LIB(installdir, pmi, function, [action-if-valid], [action-if-not-valid])
+# --------------------------------------------------------
+AC_DEFUN([OPAL_CHECK_PMI_LIB],
+[
+    OPAL_VAR_SCOPE_PUSH([save_LDFLAGS save_CPPFLAGS save_LIBS hdr_happy lib_happy files mycppflags])
+
+    # save flags
+    save_CPPFLAGS="$CPPFLAGS"
```
jsquyres, line 34: FWIW: when doing a simple assignment of one shell variable to another, you do not need quotes.  I.e.:

``` shell
save_CPPFLAGS="$CPPFLAGS"
# is functionally equivalent to (but slightly less efficient than)
save_CPPFLAGS=$CPPFLAGS
```

Unexpectedly, this is true even if the RHS contains (or may contain) whitespace.  Using quotes is slightly less efficient because it basically makes a new string and copies the value into it, and then does the assignment.

In the grand scheme of things, this minor inefficiency is irrelevant.  This is just a random point of knowledge for ya in the future.

Not asking you to change this for this PR; just a random point of knowledge for ya in the future...

ggouaillardet:
@rhc54 @jsquyres There is an issue with this commit
(-I/usr/include/slurm is not automatically set to the pmix_CFLAGS)
a suggested fix is at https://github.com/open-mpi/ompi/pull/253
in the mean time, i invite you **not** to push this commit into ompi-release

ompiteam:
Don't worry - I'm dropping this pull request anyway. I just committed an updated change to the pmi configury as the one you proposed wasn't quite right either. At this stage, I'm going back to look at the existing 1.8 code and will craft a dedicated patch as the changes in the trunk are far too specific to the situation over there.

PR #56, opened 2014-10-24
shmem/mmap: do not use O_CREAT in shared memory attach

hjelmn:
master commit
open-mpi/ompi@56a8687c2a06b503a00dc92c41bd20bf761e040d


hjelmn:
Talked to Sam about this and he confirmed it is a bug.

@jsquyres Please review

PR #59, opened 2014-10-28
oob/tcp: include a missing header file

ggouaillardet:
warning can be seen under cygwin without the missing header file

(cherry-picked from commit open-mpi/ompi@c9d1e16a9e8f4e55bbba0e1a5b7bee84594d05c1)

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp_common.c
+++ b/orte/mca/oob/tcp/oob_tcp_common.c
@@ -14,6 +14,8 @@
  * Copyright (c) 2009-2012 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2014      Intel, Inc. All rights reserved.
+ * Copyright (c) 2014      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -57,6 +59,7 @@
 #include "opal/util/argv.h"
 #include "opal/class/opal_hash_table.h"
 #include "opal/class/opal_list.h"
+#include "opal/mca/backtrace/backtrace.h"
 
 #include "orte/mca/errmgr/errmgr.h"
 #include "orte/mca/ess/ess.h"


ggouaillardet:
@rhc54 could you please review this PR ?

orte/mca/oob/tcp/oob_tcp_common.c
```diff
@@ -57,6 +59,9 @@
 #include "opal/util/argv.h"
 #include "opal/class/opal_hash_table.h"
 #include "opal/class/opal_list.h"
+#if defined(TCP_NODELAY)
+#include "opal/mca/backtrace/backtrace.h"
```
jsquyres, line 62: How is TCP_NODELAY related to opal/mca/backtrace/backtrace.h?

ggouaillardet, line 62: in orte_oob_tcp_set_socket_options() opal_backtrace_print is called only
if TCP_NODELAY is defined.
so strictly speaking, we need to include backtrace.h only if TCP_NODELAY
is defined

#if defined(TCP_NODELAY)
    int optval;
    optval = 1;
    if(setsockopt(sd, IPPROTO_TCP, TCP_NODELAY, (char *)&optval,
sizeof(optval)) < 0) {
        opal_backtrace_print(stderr, NULL, 1);
        opal_output(0, "[%s:%d] setsockopt(TCP_NODELAY) failed: %s (%d)",
                    **FILE**, **LINE**,
                    strerror(opal_socket_errno),
                    opal_socket_errno);
    }
#endif

jsquyres, line 62: Mmm.  Ok.

Could you then add a 1-line comment, perhaps something like:

``` c
#if defined(TCP_NODELAY)
/* Backtrace functionality only used if TCP_NODELAY is available */
#include "opal/mca/backtrace/backtrace.h"
...
```

jsquyres, line 62: I realize I'm being picky, but I'm sure that Future Jeff (and perhaps others?) would read that code and ask exactly the same question -- "Hmm.  How is TCP_NODELAY related to the backtrace functionality?"

:-)

ggouaillardet:
@jsquyres i removed the TCP_NODELAY reference and always include backtrace.h
i also updated the PR

jsquyres:
Good enough for me. :+1:

PR #60, opened 2014-10-28
wrappers: add the $(EXEEXT) extension to the installed symbolic links

ggouaillardet:
This is a backport of commit open-mpi/ompi@eef7590e585fbdaaadf90d9f890b32367e5320e0

@jsquyres could you please review this commit ?

Under cygwin, make install cannot be ran twice without this commit (cannot overwrite an existing symbolic link)


jsquyres:
It's ok since @jurenz already reviewed this for master.

PR #61, opened 2014-10-28
vt/wrappers: add the $(EXEEXT) extension to the installed symbolic links

ggouaillardet:
(cherry picked from commit open-mpi/ompi@6af465f12da4e1b223ddba9e4ab9e73e2010ace0)

@bertwesarg @jurenz could you please review this ?

under cygwin, make install cannot be ran twice (symlink cannot be overwritten) without this commit


jurenz:
The patch looks good to me - approved.

PR #63, opened 2014-10-29
ob1: add a destructor to mca_pml_ob1_recv_request_t

ggouaillardet:
opal_mutex_t must be OBJ_DESTRUCTed in order to avoid
a memory leak (pthread_mutex_init allocates memory under
Cygwin, so pthread_mutex_destroy is mandatory)

Thanks to Marco Atzeri for reporting this issue
(cherry picked from commit open-mpi/ompi@ed93c8787d17b470bc7c1afe80536d8cacdd862d)

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.c
@@ -16,6 +16,8 @@
  * Copyright (c) 2011-2012 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * Copyright (c) 2012      FUJITSU LIMITED.  All rights reserved.
+ * Copyright (c) 2014      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  * 
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@rhc54 could you please review this ?

jsquyres:
Let's have @bosilca and/or @hjelmn have a look...  Ralph's not much into the MPI/PML kind of stuff.

rhc54:
I agree I'm not the PML expert, but I was wondering if we could/should put a #if CYGWIN around this so we don't pickup the extra code path when it isn't required? I would think running the destructor could impact the critical path, yes?

jsquyres:
@rhc54 Yes, this is why I wanted @bosilca and/or @hjelmn to have a look.  Frankly, I'm not sure why these requests are being _destroyed_, apparently in the critical path.  To my non-PML-expert understanding, they should be getting freelisted and recycled, which makes me a little wary of this solution -- and why I think a PML expert needs to have a look at this.

bosilca:
As Jeff stated the receive requests are recycled through the recv freelist, so in theory this destructor should only be called once (during MPI_FINALIZE). Thus this destructor is outside the critical path and should have little impact other than the positive memory leak removal.

jsquyres:
Can we confirm that the destruct is actually only being called during MPI_FINALIZE?  I thought the bug reported was that the size of the program was growing without bound...?

I.e., the use of the freelist should theoretically limit the size of the app if the requests are actually being recycled properly.  Something doesn't feel right here...

jsquyres:
@bosilca I just talked to @hjelmn -- he said that the requests are on the stack, and therefore they are being manually OBJ_CONSTRUCTed and OBJ_DESTRUCTed.

Hence, this destructor _is_ being called in the critical path.

Which raises the question: what's it going to do to our performance?  On Linux, the destructor is effectively a no-op, but it still invokes at least 2 functions (the destructor and then the pthread destructor).  Does this swing the performance needle back in favor of using the freelist and not using stack-based requests?

hjelmn:
With this commit applied I see no change in the small message shared memory latency measured with osu_latency on an Opteron 6136. I will test on other systems later today.

bosilca:
This needs the open-mpi/ompi@924d39e415e7df53841803941e450aa42b119650 to always OBJ_DESTRUCT the alloca request. I tested on latency on 3 different platforms with vader and sm and I could not see any noticeable impact due to this change.

jsquyres:
Separate PR filed for George's additional commit (and then one more minor fix after that): #66

PR #64, opened 2014-10-30
Update the PMI configure logic

rhc54:
Update the PMI configure logic to the minimum level required to address user-reported problems

@jsquyres please review

### DIFF
--- a/config/opal_check_pmi.m4
+++ b/config/opal_check_pmi.m4
@@ -1,4 +1,4 @@
-# -*- shell-script -*-
+# -*- shell-script ; indent-tabs-mode:nil -*-
 #
 # Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
 #                         University Research and Technology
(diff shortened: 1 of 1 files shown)


rhc54:
FWIW: I have verified this patch on the release branch. Just waiting for official "okey-dokey" from @jsquyres

jsquyres:
@rhc54 I have a dim recollection that this PR does not directly correspond to master commits because v1.8 has diverged too far -- is that right?

If so, is that why this commit does not reflect the simpler breakdown of helper macros in opal_check_pmi.m4 from master?

rhc54:
Yeah, that's the problem all right. In master, we have separate pmi1 and pmi2 components, and so we need to reflect that in the logic. In 1.8, we only have one pmi component and use #if's to select pmi2 vs pmi1.

jsquyres:
Good enough for v1.8.

PR #66, opened 2014-10-31
ob1 destructor fix #2

jsquyres:
Minor fix to ensure that the request destructor is always called in ob1, even in error cases.

Submitted by @bosilca / @jsquyres, reviewed by @jsquyres / @bosilca.


jsquyres:
Oops, forgot to include master commits:
- open-mpi/ompi@924d39e415e7df53841803941e450aa42b119650
- open-mpi/ompi@7a5b2e9

PR #67, opened 2014-10-31
Revert most of open-mpi/ompi@6ef938de3fa9ca0fed2c5bcb0736f65b0d8803af

jsquyres:
We only intended to bring over the "remove --enable-ompi-progress-threads" stuff, but instead brought over all the stuff to enable thread locking by default (which should only be on master, not over here on the v1.8 branch).

Original commit on master: open-mpi/ompi@6ef938de3fa9ca0fed2c5bcb0736f65b0d8803a

Hence, this commit reverts all of the above-mentioned commit _except_ the --enable-ompi-progress-threads stuff (i.e., we still leave that stuff deleted from the v1.8 branch).

@hjelmn please review


hjelmn:
I don't see any gotchas in there. Lets go ahead an pull this and just watch the MTTs carefully.

PR #69, opened 2014-11-03
Don't open PSM context when run on single node

afriedle-intel:
When running many ranks (err, processes) on a single node using PSM, it's possible to
exhaust the network hardware contexts (there are 16).  This patch checks
if only a single node is being used. If so, the 'ipath' component of PSM
is disabled and no hardware contexts are opened.

This patch has been reviewed and verified internally.

(cherry picked from commit open-mpi/ompi@75f66605e568b87b4ca988ed1bfa939cd1574773)


rhc54:
Andrew confirms this has been reviewed and tested inside the Intel repo.

PR #70, opened 2014-11-03
opal_check_ident.m4: make the test match what we do in the code base

jsquyres:
Per open-mpi/ompi#257, the configure test used to check:

``` c
#pragma ident "foo"
```

But out in the code base, we actually do this:

``` c
#define IDENT_MSG "foo"
// ...
#pragma ident IDENT_MSG
```

This commit updates the configure test to `#define` a string and then use that with the various ident strategies.


jsquyres:
Issue raised by @crtrott 

@rhc54 Mind having a look?  Trivial m4 stuff.

PR #73, opened 2014-11-04
coll/basic: fix segmentation fault in neighborhood collectives if the de...

ggouaillardet:
...gree

of the topology is higher than the communicator size

It is possible to have a topology degree higher than the size of the communicator.
For example, a periodic cartesian communicator on MPI_COMM_SELF. This will leave
the neighborhood collectives with a request buffer that is too small. This commit
adds a call that will dynamically increase the size of the request buffer if it
is too small.

A better fix would be to create the topology _before_ calling the coll_select
routine on a communicator. This will take some discussion and the solution will
not likely be ready anytime soon.

Thanks to Lisandro Dalcin for reporting this.

Original thread: http://www.open-mpi.org/community/lists/devel/2014/08/15713.php

cmr=v1.8.3:reviewer=jsquyres

This commit was SVN r32796.
(cherry picked from commit 9c788ff9400960823637dc0eebb3a8640414fa05)

### DIFF
--- a/ompi/mca/coll/basic/coll_basic.h
+++ b/ompi/mca/coll/basic/coll_basic.h
@@ -12,7 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2008      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2012      Sandia National Laboratories. All rights reserved.
- * Copyright (c) 2013      Los Alamos National Security, LLC. All rights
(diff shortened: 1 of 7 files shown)


ggouaillardet:
@rhc54 this is a PR from @hjelmn commit to the master.
we will likely end up doing it differently on the master, but we'll stick to this for the v1.8 branch.
i reviewed this and this is ok for me.

PR #74, opened 2014-11-04
fortran: fix linking with Fortran sentinel values

jsquyres:
In OMPI 1.8 and 1.8.1, we didn't allow gfortran to use the ignore-TKR style "mpi" Fortran module (which is written in Fortran).  Instead, gfortran used the TKR style "mpi" module, which is written in C.

In OMPI 1.8.2, we removed the gfortran restriction (for suitably modern gfortran -- i.e., >= 4.9).  This allowed gfortran to use the ignore-TKR style "mpi" module, but exposed a bug where MPI Fortran sentinel values wouldn't be linked properly on OS X.  The solution is to use `-Wl,-flat_namespace` (see the OS X ld(1) man page for details) when both building the libraries and in the wrapper compilers.

Thanks to Github user @yafshar for raising the issue.

Fixes open-mpi/ompi#259.

(cherry picked from commit open-mpi/ompi@6164dc4960805b5502f86f9a2dccc5fc557e44d6)

### DIFF
--- a/config/ompi_setup_fc.m4
+++ b/config/ompi_setup_fc.m4
@@ -13,7 +13,7 @@ dnl                         All rights reserved.
 dnl Copyright (c) 2007      Los Alamos National Security, LLC.  All rights
 dnl                         reserved. 
 dnl Copyright (c) 2007      Sun Microsystems, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


jsquyres:
@goodell Can you have a look?

goodell:
Looks OK to me.

PR #84, opened 2014-11-07
Make sure CUDA init called during BTL init

rolfv:
Need to ensure that CUDA init is called during init to match corresponding finalize during closing.
open-mpi/ompi:022612c83ba541678015791fab1a51b7c8ce9fb4
open-mpi/ompi:cbb43d5ac3e5fa96a001097f2b4c680ffef74798

@jsquyres can you review?

### DIFF
--- a/ompi/mca/btl/sm/btl_sm_component.c
+++ b/ompi/mca/btl/sm/btl_sm_component.c
@@ -732,6 +732,10 @@ mca_btl_sm_component_init(int *num_btls,
     mca_btl_sm_component.sm_mpool = NULL;
     mca_btl_sm_component.sm_mpool_base = NULL;
 
+#if OPAL_CUDA_SUPPORT
+    mca_common_cuda_stage_one_init();
+#endif /* OPAL_CUDA_SUPPORT */
+
     /* if no session directory was created, then we cannot be used */
     if (NULL == ompi_process_info.job_session_dir) {
     /* SKG - this isn't true anymore. Some backing facilities don't require a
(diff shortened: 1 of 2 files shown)


mellanox-github:
Merged build triggered.

mellanox-github:
Merged build started.

mellanox-github:
Merged build finished. Test PASSed.

mellanox-github:
Test PASSed.
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/14/
Test PASSed.

jsquyres:
I can see that the stage one init was moved up, but I don't really have any basis to know why it was moved up, or what the side-effects might be...?

The commit message isn't very accurate or helpful.  :-(

@rolfv Can you explain further?

rolfv:
The idea is that any BTL that may support CUDA GPU buffers needs to make sure it does an initialization during MPI_Init.  There could be multiple BTLs that support it so each one makes a call into the initialization routine.  In addition, each BTL calls into a finalization routine when it is done.  This can happen during MPI_Finalize or it can happen earliier if the BTL is not selected to be used or gets some type of error during the initialization.   The common cuda code will do the initialization when the first BTL calls into it and will do the finalization cleanup when the last one leaves.  Therefore, I need to have a match between all inits and finalizes.  Even if a BTL get some type of failure during init, it still calls finalize.  So, I have moved the code init earlier in the routine so I have a match for each call to finalize.

jsquyres:
Got it.  A little more verbose explanation in the commit message would be appreciated next time.  :-)

PR #102, opened 2014-12-01
lsf configury: add dependent libraries for static linking

jsquyres:
Ensure to add the LSF dependent libraries and LD flags for the wrapper compiler static linking case.

Note that we add the same flags in all cases, but opal_mca.m4 will de-duplicate the flags so that they aren't actually added to the wrapper linker flags multiple times.

(cherry picked from commit open-mpi/ompi@56cfa90ddaaf9939926ff54b5f2f34e32254681e)

This issue was brought up as a side-effect of https://github.com/open-mpi/ompi/issues/271.

@rhc54 can you review?


mellanox-github:
Test PASSed.
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/40/
Test PASSed.

rhc54:
Question: if we had to do this for LSF, then don't we need to do it for the others (e.g., doesn't tm need to add Torque, etc.)?

jsquyres:
@rhc54 Yes.  Are you volunteering? :-)

rhc54:
I guess so - fwiw, we have this in 1.6. Not sure why it was removed, do you have any idea?

jsquyres:
Err... sunspots?

I'm guessing this is fallout from when we added the libs_static field to the wrapper compiler data files.  I.e., this is a consequence that wasn't envisioned.

rhc54:
I'm going to close this PR and replace it with the complete one

PR #109, opened 2014-12-05
Shorten the timeout period when we attempt to create a connection

rhc54:
Add an MCA parameter to allow it to be tuned by users and admins, but default it to 0.1 seconds. Need verification by others to ensure this is adequate.

@ggouaillardet please verify this and check the default value works for you

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp_connection.c
+++ b/orte/mca/oob/tcp/oob_tcp_connection.c
@@ -149,7 +149,8 @@ void mca_oob_tcp_peer_try_connect(int fd, short args, void *cbdata)
     char *host;
     mca_oob_tcp_send_t *snd;
     bool connected = false;
-
+    fd_set set;
+    
     opal_output_verbose(OOB_TCP_DEBUG_CONNECT, orte_oob_base_framework.framework_output,
                         "%s orte_tcp_peer_try_connect: "
                         "attempting to connect to proc %s",
@@ -208,20 +209,35 @@ void mca_oob_tcp_peer_try_connect(int fd, short args, void *cbdata)
         peer->active_addr = addr;  // record the one we are using
     retry_connect:
         addr->retries++;
+        FD_ZERO(&set);
+        FD_SET(peer->sd, &set);
         if (connect(peer->sd, (struct sockaddr*)&addr->addr, addrlen) < 0) {
             /* non-blocking so wait for completion */
             if (opal_socket_errno == EINPROGRESS || opal_socket_errno == EWOULDBLOCK) {
                 opal_output_verbose(OOB_TCP_DEBUG_CONNECT, orte_oob_base_framework.framework_output,
-                                    "%s waiting for connect completion to %s - activating send event",
+                                    "%s waiting %d:%03d for connect completion to %s",
+                                    ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
+                                    (int)mca_oob_tcp_component.connect_timeout.tv_sec,
+                                    (int)mca_oob_tcp_component.connect_timeout.tv_usec,
+                                    ORTE_NAME_PRINT(&peer->name));
+                /* do a select to wait */
+                if (0 > select(peer->sd + 1, &set, NULL, NULL, &mca_oob_tcp_component.connect_timeout)) {
+                    /* didn't connect, let's move to next option */
+                    addr->state = MCA_OOB_TCP_FAILED;
+                    opal_output_verbose(OOB_TCP_DEBUG_CONNECT, orte_oob_base_framework.framework_output,
+                                        "%s connection to %s failed - moving to next option",
+                                        ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
+                                        ORTE_NAME_PRINT(&peer->name));
+                    continue;
+                }
+                /* if we got a positive response, then we connected */
+                opal_output_verbose(OOB_TCP_DEBUG_CONNECT, orte_oob_base_framework.framework_output,
+                                    "%s select for connection to %s succeeded",
                                     ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
                                     ORTE_NAME_PRINT(&peer->name));
-                /* just ensure the send_event is active */
-                if (!peer->send_ev_active) {
-                    opal_event_add(&peer->send_event, 0);
-                    peer->send_ev_active = true;
-                }
-                OBJ_RELEASE(op);
-                return;
+                addr->retries = 0;
+                connected = true;
+                break;
             }
 
             /* Some kernels (Linux 2.6) will automatically software
--- a/orte/mca/oob/tcp/oob_tcp_connection.c
+++ b/orte/mca/oob/tcp/oob_tcp_connection.c
@@ -221,7 +221,8 @@ void mca_oob_tcp_peer_try_connect(int fd, short args, void *cbdata)
                                     (int)mca_oob_tcp_component.connect_timeout.tv_usec,
                                     ORTE_NAME_PRINT(&peer->name));
                 /* do a select to wait */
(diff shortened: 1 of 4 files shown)


mellanox-github:
Test PASSed.
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/50/
Test PASSed.

rhc54:
@ggouaillardet can you verify this works for you? I think the timeout is kinda low, but I'm open to suggestions.

rhc54:
BTW: according to the code, the current default timeout is 0.001 seconds. I set it way down there just to see if anyone encountered issues. So far on the trunk, no issues reported, though that seems awful fast to me. We may be hitting some minimum value in the kernel stack that overrides this setting.

jsquyres:
That does seem a bit low.  I'd go 1 second at the lowest.  5 or 10 seconds might be better...?

rhc54:
Yeah, I agree - fwiw, I've been surprised that no MTT failures or questions have been filed

mellanox-github:
Test PASSed.
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/54/
Test PASSed.

orte/mca/oob/tcp/oob_tcp_connection.c
```diff
@@ -208,20 +209,35 @@ void mca_oob_tcp_peer_try_connect(int fd, short args, void *cbdata)
...
                                     ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
+                                    (int)mca_oob_tcp_component.connect_timeout.tv_sec,
+                                    (int)mca_oob_tcp_component.connect_timeout.tv_usec,
                                     ORTE_NAME_PRINT(&peer->name));
-                /* just ensure the send_event is active */
-                if (!peer->send_ev_active) {
-                    opal_event_add(&peer->send_event, 0);
-                    peer->send_ev_active = true;
+                /* do a select to wait */
+                if (0 > select(peer->sd + 1, &set, NULL, NULL, &mca_oob_tcp_component.connect_timeout)) {
```
ggouaillardet: if a timeout occurs, then select returns 0, and this case is not handled :
a simple helloworld takes several minutes before failing

ggouaillardet:
@rhc54 sorry for the late review

here is a patch that fixes the case when select returns 0 :

``` patch
diff --git a/orte/mca/oob/tcp/oob_tcp_connection.c b/orte/mca/oob/tcp/oob_tcp_connection.c
index f368218..208676c 100644
--- a/orte/mca/oob/tcp/oob_tcp_connection.c
+++ b/orte/mca/oob/tcp/oob_tcp_connection.c
@@ -221,7 +221,8 @@ void mca_oob_tcp_peer_try_connect(int fd, short args, void *cbdata)
                                     (int)mca_oob_tcp_component.connect_timeout.tv_usec,
                                     ORTE_NAME_PRINT(&peer->name));
                 /* do a select to wait */
-                if (0 > select(peer->sd + 1, &set, NULL, NULL, &mca_oob_tcp_component.connect_timeout)) {
+                rc = select(peer->sd + 1, &set, NULL, NULL, &mca_oob_tcp_component.connect_timeout);
+                if (0 > rc) {
                     /* didn't connect, let's move to next option */
                     addr->state = MCA_OOB_TCP_FAILED;
                     opal_output_verbose(OOB_TCP_DEBUG_CONNECT, orte_oob_base_framework.framework_output,
@@ -229,6 +230,17 @@ void mca_oob_tcp_peer_try_connect(int fd, short args, void *cbdata)
                                         ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
                                         ORTE_NAME_PRINT(&peer->name));
                     continue;
+                } else if (0 == rc) {
+                    if (mca_oob_tcp_component.max_retries < addr->retries) {
+                        opal_output_verbose(OOB_TCP_DEBUG_CONNECT, orte_oob_base_framework.framework_output,
+                                            "%s orte_tcp_peer_try_connect: %s:%d retries exceeded",
+                                            ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
+                                            opal_net_get_hostname((struct sockaddr*)&addr->addr),
+                                            opal_net_get_port((struct sockaddr*)&addr->addr));
+                        continue;
+                    } else {
+                        goto retry_connect;
+                    }
                 }
                 /* if we got a positive response, then we connected */
                 opal_output_verbose(OOB_TCP_DEBUG_CONNECT, orte_oob_base_framework.framework_output,
```

IMHO, a down loopback interface (at least on Linux) is a system error, and i am fine if OMPI simply reports it, so the enduser can have this fixed by the sysadmin.

here is a pretty trivial patch (to be finalized) that does this

``` patch
diff --git a/orte/mca/oob/tcp/help-oob-tcp.txt b/orte/mca/oob/tcp/help-oob-tcp.txt
index c5eda40..c17839e 100644
--- a/orte/mca/oob/tcp/help-oob-tcp.txt
+++ b/orte/mca/oob/tcp/help-oob-tcp.txt
@@ -64,3 +64,7 @@ value will be ignored.
   Local host: %s
   Value:      %s
   Message:    %s
+#
+[no-loopback-found]
+WARNING: No loopback interface was found.
+You will likely run into some trouble...
diff --git a/orte/mca/oob/tcp/oob_tcp_component.c b/orte/mca/oob/tcp/oob_tcp_component.c
index 3908a2c..96924bc 100644
--- a/orte/mca/oob/tcp/oob_tcp_component.c
+++ b/orte/mca/oob/tcp/oob_tcp_component.c
@@ -436,6 +436,7 @@ static bool component_available(void)
     char name[32];
     struct sockaddr_storage my_ss;
     int kindex;
+    bool loopback = false;

     opal_output_verbose(5, orte_oob_base_framework.framework_output,
                         "oob:tcp: component_available called");
@@ -459,6 +460,7 @@ static bool component_available(void)

     /* look at all available interfaces */ 
     for (i = opal_ifbegin(); i >= 0; i = opal_ifnext(i)) {
+        loopback |= opal_ifisloopback(i);
         if (OPAL_SUCCESS != opal_ifindextoaddr(i, (struct sockaddr*) &my_ss,
                                                sizeof (my_ss))) {
             opal_output (0, "oob_tcp: problems getting address for index %i (kernel index %i)\n",
@@ -559,6 +561,13 @@ static bool component_available(void)
                                 opal_net_get_hostname((struct sockaddr*) &my_ss));
         }
     }
+    if (!loopback) {
+        /* FIXME
+         * display a warning only on the relevant arch
+         * for example on Linux but not on Solaris
+         */
+        orte_show_help("help-oob-tcp.txt", "no-loopback-found", true);
+    }

     /* cleanup */
     if (NULL != interfaces) {
```

rhc54:
@ggouaillardet I really appreciate your suggestion - no worries about the response time (pardon the pun). My only concern with it is that we will still "hang" for a couple of minutes before finally exiting, ableit with a better warning message. I'm okay with that for the loopback issue as that is very rare, but it doesn't fully address the overall problem reported by Reuti (for example) of long times taken for us to failover across connection options before finally hitting the correct one and connecting.

We intend to resolve that in another way on the trunk (using a better reachability algo) - do you feel it would be okay to leave it unresolved for 1.8? In other words, is it enough to resolve just the missing loopback issue, or do we need to also address the overall connection time?

Note that our current suggestion for the latter is to tell people to use the oob_tcp_if_include/exclude options. I'm fine with it, but we did get some grumbling.

rhc54:
@ggouaillardet I updated the PR based on your comments. As stated in the latest commit, I decided to keep the timeout adjustment as well as the warning.

Please let me know what you think.

mellanox-github:
Test PASSed.
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/55/
Test PASSed.

rhc54:
I'm going to take this one for now as it is passing all my tests, pending raised issues.

PR #110, opened 2014-12-09
opal_progress: fix sched_yield check

hjelmn:
master commit open-mpi/ompi@d0da29351f9cf070abd4b6ee8409780ff02bb4b6

### DIFF
--- a/opal/runtime/opal_progress.c
+++ b/opal/runtime/opal_progress.c
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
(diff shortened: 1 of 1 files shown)


hjelmn:
@rhc54 Please review.

HAVE_SCHED_YIELD is never defined. Instead OPAL_HAVE_SCHED_YIELD is set to 0 or 1 depending on whether the call is available. Not sure when this changed but the code using sched_yield () was never updated.

jsquyres:
I think you meant @rhc54, not @rhc.  :-)

hjelmn:
Ah, yes :)

mellanox-github:
Test PASSed.
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/51/
Test PASSed.

rhc54:
Looks fine. However, I would like to wait and see about switching to use nanosleep as we know that sched_yield doesn't do anything in typical environments. We have a precedence for doing nanosleep, including using it in MPI_Finalize and throughout orte as a replacement for sched_yield when we first discovered this issue a couple of years ago. Not sure why it didn't get replaced here as well - just an oversight?

hjelmn:
Probably was an oversight.

I am looking for comments on an RFC to add an opal abstraction around nanosleep. If that makes it into trunk I will backport it to 1.8 (minus the fifo/lifo changes).

PR #115, opened 2014-12-12
Don't install the Java shmem tools if we aren't building Java support.

rhc54:
Thanks to Paul Hargrove for noticing.

Taken from open-mpi/ompi@064a241145d7384ca8d759c5ef41261d0f242e94

@jsquyres please comment


oshmem/tools/wrappers/Makefile.am
```diff
@@ -24,8 +24,10 @@ install-exec-hook:
 	(cd $(DESTDIR)$(bindir); rm -f oshcc$(EXEEXT); $(LN_S) mpicc$(EXEEXT) oshcc$(EXEEXT))
 	(cd $(DESTDIR)$(bindir); rm -f shmemfort$(EXEEXT); $(LN_S) mpifort$(EXEEXT) shmemfort$(EXEEXT))
 	(cd $(DESTDIR)$(bindir); rm -f oshfort$(EXEEXT); $(LN_S) mpifort$(EXEEXT) oshfort$(EXEEXT))
+if OMPI_WANT_JAVA_BINDINGS
```
jsquyres, line 27: I think this should be "#if OSHMEM_WANT_JAVA_BINDINGS".

rhc54, line 27: Hmmm....forgot there is a separate flag for oshmem java. Sigh. Will fix

mellanox-github:
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/58/
Test PASSed.

oshmem/tools/wrappers/Makefile.am
goodell, line 27: Why would it be `#if` instead of `if`?  This is an automake conditional, right?

http://www.gnu.org/software/automake/manual/html_node/Usage-of-Conditionals.html#Usage-of-Conditionals

rhc54, line 27: It isn’t - the issue was that there are no shmem java bindings in 1.8, and so that conditional is never defined

jsquyres, line 27: Yeah, and I typed too quickly while I was rushing to catch a plane. No # intended.

PR #120, opened 2014-12-13
This will be the official pull request to fix thread-multiple on 1.8.4

rhc54:
Had merge conflicts on the other one - not entirely sure why, but it appears that maybe others were pushing things into it, and the branch got out-of-sync. Anyway, this has the complete and cleaned-up set of commits.


mellanox-github:
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/66/
Test PASSed.

mike-dubman:
@rhc54 - the threads test did not run on this pull request.
(once it was automatically if word "thread" was part of branch name, now it should be specifically requested by adding "jenkins: thread" string into project title.
Could you please add it to the PR title?
Thanks

mike-dubman:
@rhc54 , @bosilca - I tried this patch and it still hangs with tcp:

``` shell
+ timeout -s SIGKILL 13m mpirun -np 4 -bind-to core -mca pml ob1 -mca btl self,tcp /labhome/miked/workspace/git/mellanox-hpc/ompi-release/ompi_install1/thread_tests/thread-tests-1.1/overlap 8
[warn] opal_libevent2021_event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] opal_libevent2021_event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] opal_libevent2021_event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] opal_libevent2021_event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
Time per iteration on each process (ms)
Time     Compute time    Comm time
...hang...
```

bosilca:
There are good news and less good news. The good part: with the last push (commits from open-mpi/ompi@e20413c88589d0d73147c917d240b556ea9ff6f7 till open-mpi/ompi@2fec570fe7236b76f6d4fe76a14af51b907781f4) things are extremely clean at the TCP level. Unfortunately this was not enough to solve all deadlocks and finding the culprit was an extremely laborious operation.

I noticed that a very common way to hit a deadlock was when 2 threads blocking each other deep in libevent. After a little digging, it turns out that event_process_active_single_queue is not as safe as expected, as it sets the current_event_waiters to zero with complete disregard for the previous value. As a result, a thread waiting on the condition inside del_event_internal will remain blocked if another threads hits the event_process_active_single_queue (this is possible as the thread blocked in the condition released the mutex). As the expected number of waiters is now reset to zero, the associated condition will never be raised and the deadlock occurs.

Based on this I hoped that commenting the line 1354
`base->current_event_waiters = 0;`
will be enough to remove the deadlocks, but it turns out I was too optimistic. The same type of deadlock continue to seldom happens, and realistically I don't have the time to dig into it more. Thus, I took a more defensive stance and I think I have a 100% thread-safe solution. This solution enforce the already in-place protection that prevents multiple threads from dealing with the same event base. It would not be unexpected if this change has an impact on performance, especially at scale, but we gain in safety.

Here is the extra patch that needs to be applied. @hjelmn and @jsquyres please comment, and if we all agree feel free to push it.

```
diff --git a/opal/mca/event/libevent2021/libevent/event.c b/opal/mca/event/libevent2021/libevent/event.c
index 1171c81..7991d8c 100644
--- a/opal/mca/event/libevent2021/libevent/event.c
+++ b/opal/mca/event/libevent2021/libevent/event.c
@@ -1566,14 +1566,16 @@ event_base_loop(struct event_base *base, int flags)
         * as we invoke user callbacks. */
        EVBASE_ACQUIRE_LOCK(base, th_base_lock);

-       if (base->running_loop && 0 == warn_once) {
+       if (base->running_loop) {
+            if (0 == warn_once) {
                event_warnx("%s: reentrant invocation.  Only one event_base_loop"
                    " can run on each event_base at once.", __func__);
                 /********* OMPI CHANGE **********/ 
                 warn_once = 1; 
                 /*********/
-                EVBASE_RELEASE_LOCK(base, th_base_lock);
-               return -1;
+            }
+            EVBASE_RELEASE_LOCK(base, th_base_lock);
+           return -1;
        }

        base->running_loop = 1;
```

rhc54:
FWIW: this will be fine at the ORTE level. We always have only one thread
calling event_base_loop on any given event_base, so we are not impacted by
this change.

hjelmn:
@bosilca +1 from me

bosilca:
open-mpi/ompi@357daa834e071f63471e44f047c445cc4d2c9663

bosilca:
Patch available at [gist](https://gist.github.com/0df9df41da72883addad).

mike-dubman:
currently, following btls failing with threads: openib, tcp,vader

hjelmn:
@miked-mellanox vader should be thread safe on all platforms. can you point me at a failure?

mike-dubman:
- the following command fails

``` shell
mpirun -np 2 -bind-to core -mca pml ob1 -mca btl self,vader latency_th 8
```
- you can ask jenkins to try it by the following (will update wiki soon)

*\* change PR subject to contain string: 

``` shell
jenkins: threads, known_issues
```

*\* add comment to PR: 

``` shell
retest this
```

bosilca:
After Aurelien fixed the issue with vader (which didn't work on any box without yama), I can run the test for quite some time before it complains about a truncated message.

```
[dancer01:8637] *** An error occurred in MPI_Recv
[dancer01:8637] *** reported by process [3938189313,140084653326337]
[dancer01:8637] *** on communicator MPI_COMM_WORLD
[dancer01:8637] *** MPI_ERR_TRUNCATE: message truncated
[dancer01:8637] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[dancer01:8637] ***    and potentially your MPI job)
```

As I can replicate the same behavior independent of the network I decided to take a look at the test itself. I think the test is wrong, as one can match messages with the right tag but an iteration with a larger size (generated by another thread on the remote process). One can fix it by forcing some logical ordering between the messages, either by forcing a relationship between the tag and the message size or by forcing each pair of threads to use a specific communicator.

PR #142, opened 2014-12-19
Attribute "fixes"

jsquyres:
I put "fixes" in quotes because I don't fully agree with the solution, but I don't fully disagree, either.

So pull it over, let it fix PETSc's current behavior.  But at least also update our man pages to indicate that users should not expect adding/deleting attributes to work in attribute callback functions.

I reviewed George's code; the man pages hypothetically don't need reviewing.


jsquyres:
Adding @bosilca in this issue, just so he can share the email load.  :-)

mellanox-github:
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/92/
Test PASSed.

PR #157, opened 2015-01-06
r2: fix show_help message to show the right hostname

jsquyres:
This is a fix of a fairly obvious copy-n-paste error in the arguments to show_help: it was showing the same local hostname twice, rather than the local hostname and the remote hostname. 

This problem does not exist on master, and there is therefore no corresponding master commit.

Reviewed by @goodell


mellanox-github:
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/106/
Test PASSed.

PR #163, opened 2015-01-09
btl/tcp: fix overflow in mca_btl_tcp_endpoint_dump()

ggouaillardet:
```
     and turn OFF the TCP dump mechanism
```

backported from commits :
- open-mpi/ompi@8ab605d9c5bc82ef2baf424a5d5f970132d89ca1
- open-mpi/ompi@8ddd3b3b0998f5de66bd4d392789af345eaae0e5
- open-mpi/ompi@1344097d35f44b64a35b8d97c95033a350a5cb6a

### DIFF
--- a/ompi/mca/btl/tcp/btl_tcp_frag.c
+++ b/ompi/mca/btl/tcp/btl_tcp_frag.c
@@ -10,6 +10,8 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2008-2012 Oracle and/or all its affiliates.  All rights reserved.
+ * Copyright (c) 2015      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@bosilca could you please review this ?

@rolfv this PR is for issue open-mpi/ompi#338

mellanox-github:
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/120/
Test PASSed.

bosilca:
It won't be very objective to ask me to review it as I partially wrote the patch. Let's find another volunteer ;)

@goodell can you give us a quick review on this please?

ompi/mca/btl/tcp/btl_tcp_frag.c
```diff
@@ -99,14 +101,16 @@ OBJ_CLASS_INSTANCE(
...
 size_t mca_btl_tcp_frag_dump(mca_btl_tcp_frag_t* frag, char* msg, char* buf, size_t length)
 {
-    int i, used = 0;
+    int i, used;
 
-    used += snprintf(&buf[used], length - used, "%s frag %p iov_cnt %d iov_idx %d size %lu\n",
-                     msg, (void*)frag, (int)frag->iov_cnt, (int)frag->iov_idx, frag->size);
+    used = snprintf(buf, length, "%s frag %p iov_cnt %d iov_idx %d size %lu\n",
+                    msg, (void*)frag, (int)frag->iov_cnt, (int)frag->iov_idx, frag->size);
+    if (used >= length) return length;
```
goodell, line 108: You may want to follow the pattern used in the `out:` label in `mca_btl_tcp_endpoint_dump` and manually terminate the string to be completely safe.  IIRC, some versions of `snprintf` are buggy and won't reliably `'\0'`-terminate a string when the length parameter is exceeded.

ompi/mca/btl/tcp/btl_tcp_frag.c
```diff
@@ -99,14 +101,16 @@ OBJ_CLASS_INSTANCE(
...
     for( i = 0; i < (int)frag->iov_cnt; i++ ) {
         used += snprintf(&buf[used], length - used, "[%s%p:%lu] ",
                          (i < (int)frag->iov_idx ? "*" : ""),
                          frag->iov[i].iov_base, frag->iov[i].iov_len);
+        if (used >= length) return length;
```
goodell, line 113: ditto here (easily solved with a common `out:` label)

goodell:
In general it looks pretty good.  I put a small line comment on something you might want to improve, but it's not a huge issue.  A truly pedantic review would be concerned about snprintf returning a negative error, possibly throwing off some of your `used` computation logic, but I don't think that's something we need to be concerned about in practice.

rhc54:
Just for my clarification: this not only cleans up the dump code, but it also turns that code "off" unless someone alters the source to change the #define. Is that correct?

I have no issue approving it either way, but just want to understand the full impact of the change. Is this degree of change of concern to you folks - i.e., if a user does something where dump has been called before, will we hear a squawk because it now is silent?

bosilca:
Yes WANT_PEER_DUMP completely disable the dump code. This is heavy duty debugging and should not be enabled for users. The TCP BTL has the usual opal_output_verbose support, and this should hopefully be enough for all users needs.

PR #164, opened 2015-01-09
Fix CUDA registering of shared memory. 

rolfv:
Now we register the entire region.  Otherwise, we were only registering our local pieces and this gave bad performance.

Backported from commits: 
- https://github.com/open-mpi/ompi/commit/f55de452abdeeaa282205b21f6183e3e09cfdeba
- https://github.com/open-mpi/ompi/commit/26482db7366a8f544d429ffcc800ae5e23b7e95b

@bosilca Could you review this George?  The change here is that I am registering the entire mapped memory block with CUDA so copying into and out of the memory is faster.  Without this change, I only mapped in the memory that was locally allocated within the region.  This has been in the master for about a month.


bosilca:
:+1:

mellanox-github:
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/121/
Test PASSed.

PR #179, opened 2015-02-05
Fix singleton comm_spawn

rhc54:
There were a few bugs in the singleton comm_spawn procedure, mostly caused by a typo in the novm state machine, the need to use the direct routed component, and the need to pass any prefix required for subsequent launch of daemons.

@jsquyres Don't know anyone else who might be able to review this, so I would appreciate your help.


mellanox-github:
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/134/
Test PASSed.

jsquyres:
What was the use case that didn't work?

I tried the orte/test/mpi/simple_spawn program as a singleton, both on the actual v1.8 branch and on this PR branch.  It worked in both cases...?

rhc54:
Did you try it in the case where the spawn occurs across multiple nodes? Ensure you have more than one node in your allocation/hostfile, then add "OMPI_MCA_rmaps_base_mapping_policy=node" to your environment.

PR #294, opened 2015-05-27
configury: manually remove some stamp-h? files

jsquyres:
Due to what might be a bug in Automake, we need to remove stamp-h? files manually.  See http://debbugs.gnu.org/cgi/bugreport.cgi?bug=19418.

(cherry picked from commit open-mpi/ompi@40dd4c5b766ff62a681692b1fa6b72a1023fc81f) 

This was fixed long ago on master; it wasn't brought over to v1.8 (and therefore v1.10) because we didn't have embedded libfabric on v1.8(v1.10), and therefore didn't trip this (assumed) Automake bug.  Without this PR, "make distcheck" fails because some stamp-h? files are left in the build tree.

### DIFF
--- a/opal/mca/hwloc/hwloc191/Makefile.am
+++ b/opal/mca/hwloc/hwloc191/Makefile.am
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2011-2013 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2011-2014 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2014      Intel, Inc. All right reserved.
 # $COPYRIGHT$
 # 
@@ -7,6 +7,13 @@
 # 
 # $HEADER$
 #
+
+# Due to what might be a bug in Automake, we need to remove stamp-h?
+# files manually.  See
+# http://debbugs.gnu.org/cgi/bugreport.cgi?bug=19418.
+DISTCLEANFILES = \
+    hwloc/include/hwloc/autogen/stamp-h? \
+    hwloc/include/private/autogen/stamp-h?
 
 # Need to include these files so that these directories are carried in
 # the tarball (in case someone invokes autogen.sh on a dist tarball).


opal/mca/hwloc/hwloc191/Makefile.am
```diff
@@ -8,6 +8,13 @@
 # $HEADER$
 #
 
+# Due to what might be a bug in Automake, we need to remove stamp-h?
+# files manually.  See
+# http://debbugs.gnu.org/cgi/bugreport.cgi?bug=19418.
+DISTCLEANFILES = \
+    hwloc/include/hwloc/autogen/stamp-h? \
+    hwloc/include/private/autogen/stamp-h?
```
goodell, line 16: Will the `?` wildcard really work here?  That's surprising to me, since automake doesn't usually permit wildcards in other areas...

mellanox-github:
- Coverity found 1472 errors for all
- Coverity found 23 errors for oshmem
- Coverity found 7 errors for fca
- Coverity report at http://bgate.mellanox.com:8888/jenkins/job/gh-ompi-release-v1.8-pr//307/Coverity_Report <br><br>:octocat: _Sent from [GH](http://nodegh.io)._

mellanox-github:
http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/307/
Test PASSed.

opal/mca/hwloc/hwloc191/Makefile.am
jsquyres, line 16: Yeah, it does -- these values are basically passed directly to "rm".

jsquyres:
I'm invoking Gatekeeper privilege to merge this now so that the nightly build can complete successfully.  @rhc54 can revert this tomorrow if he doesn't like it.

opal/mca/hwloc/hwloc191/Makefile.am
goodell, line 16: OK, but I suspect it's against the intent of `DISTCLEANFILES`, since automake probably thinks these are all explicit files and you're really passing it a shell glob expression.  If the AM folks add some additional checking in the future there's a chance we could get burned.  Leave it as-is for now, but it would be more future-proof to delete them via `distclean-local`: http://www.gnu.org/software/automake/manual/automake.html#Clean

PR #304, opened 2015-06-06
v1.8 btl_tcp_proc.c: add missing "continue"

jsquyres:
Also add another (superflous but symmetric) continue statement.

This missing "continue" statement allows IPv4 "private network" matches to fall through and allow IPv6 matches to be made -- thereby overriding the IPv4 match that was already made.

Fixes open-mpi/ompi#585 (although several of the other issues identified on open-mpi/ompi#585 still exist, the primary / initial bug that was reported there is now fixed).

(cherry picked from commit open-mpi/ompi@cddc8945e056eb05e5ad49de4578fe36f1c198a6)

@bosilca @ggouaillardet Please review.  I'll file a similar PR for v1.10.

### DIFF
--- a/ompi/mca/btl/tcp/btl_tcp_proc.c
+++ b/ompi/mca/btl/tcp/btl_tcp_proc.c
@@ -12,6 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2008-2010 Oracle and/or its affiliates.  All rights reserved
  * Copyright (c) 2013      Intel, Inc. All rights reserved
+ * Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
  * $COPYRIGHT$
  * 
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


mellanox-github:
- Coverity found 1464 errors for all
- Coverity found 23 errors for oshmem
- Coverity found 7 errors for fca
- Coverity report at http://bgate.mellanox.com:8888/jenkins/job/gh-ompi-release-v1.8-pr//322/Coverity_Report <br><br>:octocat: _Sent from [GH](http://nodegh.io)._

ggouaillardet:
@jsquyres this PR does close the issue that was reported.
if i understand correctly, it implicitly assumes a "valid" pair of IPv4 addresses is always "better" than an other "valid" pair of IPv6 addresses. previously, a "valid" pair of IPv6 addresses was always better than a pair of IPv4 addresses, even if the IPv6 were link-local addresses.
imho, both previous and current logic are very arguable.
that being said, and as far as i am concerned, i am fine with this (so 1.8.6 can be released) until we find a better option that covers all the issues listed in open-mpi/ompi#585
@bosilca are you good with this ?

jsquyres:
@ggouaillardet Yes, I think the missing `continue` was just a bug, not an intentional feature.  Specifically:
1. The public network checks (above the private network checks) had a `continue`
2. The CQ_PRIVATE_\* enums have lower values than the CQ_PUBLIC_\* enums, such that they would always be preferred in the numeric comparison

PR #326, opened 2015-06-17
Correct the way we handle binding to objects during comm_spawn

rhc54:
For comm-spawned jobs that are mapping/binding by a given object type, track where we are so a request doesn't overload when it shouldn't

Cherry-picked from open-mpi/ompi@869b2891c4e5d4a9c29a3ea270aab4b059525ae0

@hppritcha please review


mellanox-github:
http://bgate.mellanox.com/job/gh-ompi-release-v1.8-pr/346/

hppritcha:
I did check and this fixed the problem I was seeing on edison with cpi-worker.

PR #354, opened 2015-07-01
Fix PGI compilation of opal_condition test

hjelmn:
PGI was failing to build this test due to the {} initializers of some
of the global variables. The initializers were added to remove common
symbols from the test repository. This commit removes the initializers
and makes the variables static which should fix both the compile issue
and avoid the common symbols.

This commit additionally removes the check for
OPAL_ENABLE_MULTI_THREADS since Open MPI requires thread support.

### DIFF
--- a/test/threads/opal_condition.c
+++ b/test/threads/opal_condition.c
@@ -29,25 +29,9 @@
 #include "opal/threads/condition.h"
 #include "opal/sys/atomic.h"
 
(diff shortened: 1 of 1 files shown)


jsquyres:
Please assign a milestone and a reviewer.

hppritcha:
bot:retest

PR #356, opened 2015-07-01
usnic fix and minor updates

jsquyres:
Three commits:
1. Fix usnic static/disable-dlopen builds with libfabric (by using the right flags in Makefile.am).  There is no corresponding master commit, because the v1.10 usnic BTL builds directly against libfabric, whereas master builds against a common component libfabric.
2. Remove a stale header file.  This file was accidentally left over from when we converted from verbs to libfabric.  There is no corresponding master commit (master usnic switched from verbs to libfabric long ago).
3. Remove `#if HAVE_STRING_H`, which was done on master as part of a configury cleanup.  Making the usnic code match here on v1.10 makes the usnic BTL .c and .h code identical between master and v1.10.


jsquyres:
@rhc54 Other Cisco people are away on vacation for the long weekend -- could you review?  These three commits are actually pretty trivial.

bot:assign:rhc54

jsquyres:
@rhc54 Forgot to say: this PR will fix the usnic v1.10 MTT build failures.

rhc54:
:+1:

PR #360, opened 2015-07-02
libfabric: fix wrappers for static builds

jsquyres:
Need to set the WRAPPER_EXTRA flags so that the wrappers for static builds pull in -lfabric.

Also update/fix some comments.

(cherry picked from commit open-mpi/ompi@f1353947ffd0f1f395e3b90ce33fe28eb60f105f)

@yburette Please review


yburette:
Verified with PSM provider :+1:

PR #362, opened 2015-07-03
Topic/v1.8/osc zero size

ggouaillardet:


ggouaillardet:
this should be for the v1.8 branch .... closing now

PR #364, opened 2015-07-05
Protect the oob/tcp component from port scanners

rhc54:
(cherry picked from commit open-mpi/ompi@4352123c26346de6a3dbe02dafaf1e4bc432bbdb)

@jsquyres please review

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp_connection.c
+++ b/orte/mca/oob/tcp/oob_tcp_connection.c
@@ -972,7 +972,11 @@ static bool tcp_peer_recv_blocking(mca_oob_tcp_peer_t* peer, int sd,
             if (opal_socket_errno != EINTR && 
                 opal_socket_errno != EAGAIN && 
                 opal_socket_errno != EWOULDBLOCK) {
-                if (peer->state == MCA_OOB_TCP_CONNECT_ACK) {
+                if (NULL == peer) {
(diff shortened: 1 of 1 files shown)


jsquyres:
@rhc54 (and maybe @bosilca ?) Just so I understand this: is the intent that port scanners will hang up before sending the required number of bytes for the incoming blob?

rhc54:
@jsquyres Port scanners touch the port and thus cause a connect/accept to begin, but they send no bytes - they just immediately disconnect. We didn't protect against that, and so when we see the connection error, we have peer=NULL and go into the error code - which segfaults.

Reported by two independent sets of folks, hence the fix.

jsquyres:
Ah, ok.

I think this is a good fix -- it definitely fixes the `peer==NULL` case before trying to access `peer->state`, but I wonder a port scanner hanging up would end up in that code path to begin with.  Premature hangups should be detected in the block before this one (i.e., the block checking for `retval==0`).  This block should only be entered when retval is some uncommon _error_.

Shrug.  The fix is good, regardless, so it shouldn't be held up even if we don't understand how the code execution got to this specific code path from a port scanner...

:+1:

hppritcha:
bot:retest

hppritcha:
looks like we also need a PR for v2.x for this fix.
@rhc54

PR #365, opened 2015-07-05
Protect the oob/tcp component from port scanners

rhc54:
(cherry picked from commit open-mpi/ompi@4352123c26346de6a3dbe02dafaf1e4bc432bbdb)

@jsquyres same as for 1.10


jsquyres:
This is essentially the same as for #364 -- following up there.

jsquyres:
:+1:

PR #372, opened 2015-07-07
v2.x: mpi_f08: move the f08 version of buffer_detach to use-mpi-f08

jsquyres:
@ggouaillardet was right -- we should have put the ompi_buffer_detach_f08() function in the use-mpi-f08 directory to begin with.  Putting it in the mpif-h directory made it complicated as to whether the function would be built or not (e.g., whether weak symbols were supported or not, whether the profiling layer was  disabled or not, ...etc.).

Just put it in the use-mpi-f08 directory and always build it (when the mpi_f08 module is built, of course), and keep it simple.

(cherry picked from commit open-mpi/ompi@8ff60c47cc9b576346dc5c0a0ca214f813f92439)

@ggouaillardet please review


jsquyres:
@ggouaillardet reviewed on #373.

PR #373, opened 2015-07-07
v1.8: mpi_f08: move the f08 version of buffer_detach to use-mpi-f08

jsquyres:
@ggouaillardet was right -- we should have put the ompi_buffer_detach_f08() function in the use-mpi-f08 directory to begin with.  Putting it in the mpif-h directory made it complicated as to whether the function would be built or not (e.g., whether weak symbols were supported or not, whether the profiling layer was disabled or not, ...etc.).

Just put it in the use-mpi-f08 directory and always build it (when the mpi_f08 module is built, of course), and keep it simple.

(cherry picked from commit open-mpi/ompi@8ff60c47cc9b576346dc5c0a0ca214f813f92439)

@ggouaillardet please review

### DIFF
--- a/ompi/mpi/fortran/use-mpi-f08/buffer_detach.c
+++ b/ompi/mpi/fortran/use-mpi-f08/buffer_detach.c
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
+ *                         University Research and Technology
+ *                         Corporation.  All rights reserved.
+ * Copyright (c) 2004-2005 The University of Tennessee and The University
+ *                         of Tennessee Research Foundation.  All rights
+ *                         reserved.
+ * Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
+ *                         University of Stuttgart.  All rights reserved.
+ * Copyright (c) 2004-2005 The Regents of the University of California.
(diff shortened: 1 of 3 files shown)


ompi/mpi/fortran/use-mpi-f08/buffer_detach.c
```diff
@@ -0,0 +1,65 @@
...
+{
+    int c_ierr;
+    void *dummy;
+
+    c_ierr = MPI_Buffer_detach(&dummy, size);
+    if (NULL != ierr) {
+        *ierr = OMPI_INT_2_FINT(c_ierr);
+    }
+
+    if (MPI_SUCCESS == c_ierr) {
```
ggouaillardet, line 63: on second thought, and for the sake of symmetry, should we use an

```
int c_size;
```

local variable and set *size only if MPI_SUCCESS == c_ierr ?

jsquyres, line 63: I don't think it really matters...?

jsquyres, line 63: Crud; yes it does (sizeof(MPI_Fint) may not == sizeof(int)).  I'll fix.

jsquyres:
Done.

PR #387, opened 2015-07-10
VERSION: .so version number update

jsquyres:
- libmpi changed
- libmpifh changed
- libmpi_usempif08 changed
- libopen-rte changed
- libopen-pal changed

Need someone to review.


rhc54:
I think this looks okay....not an expert.

:+1:

PR #399, opened 2015-07-14
Make the connect operation in TCP be blocking to eliminate race conditions

rhc54:
Under review by reporting user


rhc54:
Verified by user

PR #400, opened 2015-07-14
fix the logic for setting stripe size and stripe count in the lustre …

edgargabriel:
…fs module. Takes now also the MPI_Info object into consideration.

Conflicts:
    ompi/mca/fs/lustre/fs_lustre_file_open.c

### DIFF
--- a/ompi/mca/fs/lustre/fs_lustre_component.c
+++ b/ompi/mca/fs/lustre/fs_lustre_component.c
@@ -43,7 +43,7 @@ int mca_fs_lustre_priority = 20;
    to 64KB. MCA parameter
    Can be changed at 
    runtime also*/
-int mca_fs_lustre_stripe_size = 1048576;
+int mca_fs_lustre_stripe_size = 0;
 int mca_fs_lustre_stripe_width = 0;
 /*
  * Instantiate the public struct with all of our public information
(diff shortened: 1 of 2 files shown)


hppritcha:
@edgargabriel do we have a test that I could run to check out this feature?

edgargabriel:
the way I tested it was to take a simple MPI_File I/O test which creates a new file, run on it lustre. If you set the mca parameter e.g. 

mpirun --mca fs_lustre_stripe_size 8 -np x ./mytest

and verify with

lfs getstripe 'name of output file' 

you should see the the file has the stripe size that you set. I tested it on the Lustre Cray system in Stuttgart, and it seemed to work. If you want, I can provide a trivial MPI file I/O test code for that.

hppritcha:
simple enough.  I'll try that approach.  Odd that IU jenkins tests got triggered by my comment.

edgargabriel:
maybe I should add a testcase to my testsuite, where we do that through an info object. That could be done semi-automatic in that case.

edgargabriel:
btw, the real problem that this commit fixes was, that the previous code version did not use the system defaults if the user did not specify the parameters, we basically always overwrote the system settings. I _suspect_ that this was one of the contributors to the read performance problem that Fujitsu observed. This fix should keep now the system settings if the user does not specify anything.

hppritcha:
Edgar, works for me if I use a stripe size thats a multilpe of 64KB.  So I had to use
mpirun --mca fs_lustre_stripe_size 1048576 -np 1 ./file_open
as a test.

I'm assuming your example with '8' left out a qualifier like 'M'.

edgargabriel:
sorry, I think my example was meant to be stripe_count, where I force lustre to use 8 servers. Not the stripe size

hppritcha:
@jsquyres I'll let you approve and merge.

edgargabriel:
@hppritcha and @jsquyres I found the issue that lead to the lustre problem, and I have a fix, the vast majority of tests pass now on lustre (except for the new file_delete test, need to double check). Should I file a new pr or issue, or is there a way to reopen this one?

jsquyres:
Make a new PR; this one has already been merged, so it's effectively done.

PR #409, opened 2015-07-22
mtl: add query method to mtl components

hppritcha:
====================Forgotten PR needed for 1.10==========================

Switch to using the query/priority method for selecting
MTLs.  This switch was motivated by the fact that now
on some platforms, its possible for multiple MTLs to
be initializable, but only one MTL should be selected.

In addition, there is a complication with the PSM and
IFO (with PSM provider) MTLs owing to the fact that
they cannot both intialize the underlying PSM context,
i.e. only one call to psm_init is allowed per process.

The mxm component has not been compiled as the author
doesn't currently have access to a system with a recent
enough mxm installed to allow for a compile.

The portals4, ofi, and psm components have been checked
for compilation.  The ofi and psm components have been
checked for runtime correctness on a intel/qlogic system
with up to date PSM installed.

Bunch of impacted MTL authors:

@bureddy  @regrant @adrianreber 

Fixes open-mpi/ompi#735

### DIFF
--- a/ompi/mca/mtl/mxm/mtl_mxm_component.c
+++ b/ompi/mca/mtl/mxm/mtl_mxm_component.c
@@ -24,8 +24,11 @@
 #include <unistd.h>
 
 static int ompi_mtl_mxm_component_open(void);
+static int ompi_mtl_mxm_component_query(mca_base_module_t **module, int *priority);
 static int ompi_mtl_mxm_component_close(void);
 static int ompi_mtl_mxm_component_register(void);
+
+static int param_priority;
 
 int mca_mtl_mxm_output = -1;
 
@@ -42,14 +45,14 @@ mca_mtl_mxm_component_t mca_mtl_mxm_component = {
      */
     {
         MCA_MTL_BASE_VERSION_2_0_0,
-        "mxm", /* MCA component name */
-        OMPI_MAJOR_VERSION, /* MCA component major version */
-        OMPI_MINOR_VERSION, /* MCA component minor version */
-        OMPI_RELEASE_VERSION, /* MCA component release version */
-        ompi_mtl_mxm_component_open, /* component open */
-        ompi_mtl_mxm_component_close, /* component close */
-        NULL,
-        ompi_mtl_mxm_component_register
+        .mca_component_name = "mxm",
+        .mca_component_major_version = OMPI_MAJOR_VERSION,
+        .mca_component_minor_version = OMPI_MINOR_VERSION,
+        .mca_component_release_version = OMPI_RELEASE_VERSION,
+        .mca_open_component = ompi_mtl_mxm_component_open,
+        .mca_close_component = ompi_mtl_mxm_component_close,
+        .mca_query_component = ompi_mtl_mxm_component_query,
+        .mca_register_component_params = ompi_mtl_mxm_component_register,
     },
     {
         /* The component is not checkpoint ready */
@@ -85,6 +88,15 @@ static int ompi_mtl_mxm_component_register(void)
                                            OPAL_INFO_LVL_9,
                                            MCA_BASE_VAR_SCOPE_READONLY,
                                            &ompi_mtl_mxm.mxm_np);
+
+    param_priority = 100;
+    (void) mca_base_component_var_register (c,
+                                            "priority", "Priority of the MXM MTL component",
+                                            MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
+                                            OPAL_INFO_LVL_9,
+                                            MCA_BASE_VAR_SCOPE_READONLY,
+                                            &param_priority);
+
 
 #if MXM_API >= MXM_VERSION(3,1)
 {
@@ -203,6 +215,18 @@ static int ompi_mtl_mxm_component_open(void)
         return OPAL_ERR_NOT_AVAILABLE;
     }
 
+    return OMPI_SUCCESS;
+}
+
+static int ompi_mtl_mxm_component_query(mca_base_module_t **module, int *priority)
+{
+
+    /*
+     * if we get here it means that mxm is available so give high priority
+     */
+
+    *priority = param_priority;
+    *module = (mca_base_module_t *)&ompi_mtl_mxm.super;
     return OMPI_SUCCESS;
 }
 
--- a/ompi/mca/mtl/portals4/mtl_portals4_component.c
+++ b/ompi/mca/mtl/portals4/mtl_portals4_component.c
@@ -28,9 +28,12 @@
 #include "mtl_portals4_recv_short.h"
 #include "mtl_portals4_message.h"
 
+static int param_priority;
+
 static int ompi_mtl_portals4_component_register(void);
 static int ompi_mtl_portals4_component_open(void);
 static int ompi_mtl_portals4_component_close(void);
+static int ompi_mtl_portals4_component_query(mca_base_module_t **module, int *priority);
 static mca_mtl_base_module_t* 
 ompi_mtl_portals4_component_init(bool enable_progress_threads, 
                                  bool enable_mpi_threads);
@@ -45,14 +48,14 @@ mca_mtl_base_component_2_0_0_t mca_mtl_portals4_component = {
     {
         MCA_MTL_BASE_VERSION_2_0_0,
 
-        "portals4", /* MCA component name */
-        OMPI_MAJOR_VERSION,  /* MCA component major version */
-        OMPI_MINOR_VERSION,  /* MCA component minor version */
-        OMPI_RELEASE_VERSION,  /* MCA component release version */
-        ompi_mtl_portals4_component_open,  /* component open */
-        ompi_mtl_portals4_component_close, /* component close */
-        NULL,
-        ompi_mtl_portals4_component_register
+        .mca_component_name = "portals4",
+        .mca_component_major_version = OMPI_MAJOR_VERSION,
+        .mca_component_minor_version = OMPI_MINOR_VERSION,
+        .mca_component_release_version = OMPI_RELEASE_VERSION,
+        .mca_open_component = ompi_mtl_portals4_component_open,
+        .mca_close_component = ompi_mtl_portals4_component_close,
+        .mca_query_component = ompi_mtl_portals4_component_query,
+        .mca_register_component_params = ompi_mtl_portals4_component_register,
     },
     {
         /* The component is not checkpoint ready */
@@ -73,6 +76,14 @@ ompi_mtl_portals4_component_register(void)
 {
     mca_base_var_enum_t *new_enum;
     int ret;
+
+    param_priority = 10;
+    (void) mca_base_component_var_register (&mca_mtl_portals4_component.mtl_version,
+                                            "priority", "Priority of the Portals4 MTL component",
+                                            MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
+                                            OPAL_INFO_LVL_9,
+                                            MCA_BASE_VAR_SCOPE_READONLY,
+                                            &param_priority);
 
     ompi_mtl_portals4.eager_limit = 2 * 1024;
     (void) mca_base_component_var_register(&mca_mtl_portals4_component.mtl_version,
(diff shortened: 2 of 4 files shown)


jsquyres:
@hppritcha Ran into some compile-time errors in the MXM component:

```
18:48:24 mtl_mxm_component.c: In function 'ompi_mtl_mxm_component_register':
18:48:24 mtl_mxm_component.c:94: error: 'MCA_COMPILETIME_VER' undeclared (first use in this function)
18:48:24 mtl_mxm_component.c:94: error: (Each undeclared identifier is reported only once
18:48:24 mtl_mxm_component.c:94: error: for each function it appears in.)
18:48:24 mtl_mxm_component.c:96: error: 'MCA_BASE_VAR_TYPE_VERSION_STRING' undeclared (first use in this function)
18:48:24 mtl_mxm_component.c:113: error: 'MCA_RUNTIME_VER' undeclared (first use in this function)
18:48:24 mtl_mxm_component.c: In function 'ompi_mtl_mxm_component_query':
18:48:24 mtl_mxm_component.c:262: warning: assignment from incompatible pointer type
18:48:24 mtl_mxm.c: In function 'ompi_mtl_mxm_del_procs':
18:48:24 mtl_mxm.c:516: warning: comparison between signed and unsigned integer expressions
```

hppritcha:
botched the cross reference to commit, updating.

PHHargrove:
Cannot compile mtl:ofi with this PR because ompi/mca/mtl/ofi/mtl_ofi_component.c contains conflict markers!!

ompi/mca/mtl/ofi/mtl_ofi_component.c
```diff
@@ -87,11 +87,19 @@ ompi_mtl_ofi_component_open(void)
     return OMPI_SUCCESS;
 }
 
+<<<<<<< HEAD
```
yburette: Looks like these should be removed.

ompi/mca/mtl/portals4/mtl_portals4_component.c
```diff
@@ -50,8 +53,8 @@ mca_mtl_base_component_2_0_0_t mca_mtl_portals4_component = {
         OMPI_MINOR_VERSION,  /* MCA component minor version */
         OMPI_RELEASE_VERSION,  /* MCA component release version */
         ompi_mtl_portals4_component_open,  /* component open */
+        ompi_mtl_portals4_component_query,  /* component close */
```
PHHargrove: Note that order here is query,close vs. close,query in the other mtl_*_component.c files.
So this one is very likely wrong.
The comment is wrong too.

ompi/mca/mtl/psm/mtl_psm_component.c
```diff
@@ -59,8 +62,8 @@ mca_mtl_psm_component_t mca_mtl_psm_component = {
             OMPI_RELEASE_VERSION,  /* MCA component release version */
             ompi_mtl_psm_component_open,  /* component open */
             ompi_mtl_psm_component_close,  /* component close */
-	    NULL,
-	    ompi_mtl_psm_component_register
+            ompi_mtl_psm_component_query,  /* component close */
```
PHHargrove: Note that comment is incorrect (says "close" when should be "query")

ompi/mca/mtl/portals4/mtl_portals4_component.c
jsquyres: These assignments should be done with C99 syntax to avoid mis-ordering problems like this.

ompi/mca/mtl/psm/mtl_psm_component.c
jsquyres: Ditto -- these assignments should also be done with C99 syntax.

ompi/mca/mtl/mxm/mtl_mxm_component.c
```diff
@@ -48,7 +51,7 @@ mca_mtl_mxm_component_t mca_mtl_mxm_component = {
         OMPI_RELEASE_VERSION, /* MCA component release version */
         ompi_mtl_mxm_component_open, /* component open */
         ompi_mtl_mxm_component_close, /* component close */
-        NULL,
+        ompi_mtl_mxm_component_query, /* component query */
         ompi_mtl_mxm_component_register
```
jsquyres: Just to beat a dead horse: these should be C99 syntax, too.  :-)

PHHargrove:
A build w/ portals4 (but without ofi) shows not only the swapped initializers, but also a missing cast:

```
/global/homes/h/hargrove/GSCRATCH/OMPI/openmpi-1.10.0rc2-linux-x86_64-icc-14/openmpi-1.10.0rc2/ompi/mca/mtl/portals4/mtl_portals4_component.c(56): warning #144: a value of type "int (*)(mca_base_module_t **, int *)" cannot be used to initialize an entity of type "mca_base_close_component_1_0_0_fn_t"
          ompi_mtl_portals4_component_query,  /* component close */
          ^

/global/homes/h/hargrove/GSCRATCH/OMPI/openmpi-1.10.0rc2-linux-x86_64-icc-14/openmpi-1.10.0rc2/ompi/mca/mtl/portals4/mtl_portals4_component.c(57): warning #144: a value of type "int (*)(void)" cannot be used to initialize an entity of type "mca_base_query_component_2_0_0_fn_t"
          ompi_mtl_portals4_component_close, /* component close */
          ^

/global/homes/h/hargrove/GSCRATCH/OMPI/openmpi-1.10.0rc2-linux-x86_64-icc-14/openmpi-1.10.0rc2/ompi/mca/mtl/portals4/mtl_portals4_component.c(232): warning #556: a value of type "mca_mtl_base_module_t *" cannot be assigned to an entity of type "mca_base_module_t *"
      *module = &ompi_mtl_portals4.base;
              ^

```

By inspection I suspect `ompi_mtl_psm_component_query()` could have the same issue (but am not sure about types of `base` vs `super`), and that the unresolved conflict in `ompi_mtl_ofi_component_query()` needs to be resolved in favor of the HEAD (which has the cast).

hppritcha:
okay pushing fixes up.  Some of these - like the portals4 problem - was actually fixed in another pr in master.  Didn' realize I was not compiling the ofi mtl on my qlogic box.  That's fixed.  But the machine is very busy and I can't get a session to test.  I'll push new changes.

@yburette could you test on one of your systems?

hppritcha:
@yburette  found a bug in the ofi mtl which will need to be fixed to avoid the segfault @PHHargrove is seeing.   I'll open a PR shortly on master for you to review.

yburette:
@hppritcha I finally got some True Scale machines to try this PR. It is working for me. The PSM MTL is being selected automatically.

@PHHargrove I also tried without any parameter (apart from -n and -H) on the command line and it seems to work.

rhc54:
@yburette The key to recreating the problem is that you have to get an MTL to build, but not have it be selected at runtime - without specifying the ^mtl param so it will be opened and then closed. Otherwise, it will always be selected and the problem won't occur.

hppritcha:
the truescale machine here is less busy today so I can run stuff.  Ralph is right - for 1.10.  You have to force selection of mtl to be ofi, but because of the not-so-great things going on in the cm select method, you end up using the ob1 pml.  On master, the not-so-great things going on in the cm select method allow things to work -depending on your compiler.

Any rate, @yburette  please review open-mpi/ompi#747.

yburette:
@rhc54, I agree. I forgot to mention that both the PSM and OFI MTLs were built.

```
[f011:125594] mca: base: components_open: component psm open function successful
[f011:125594] mca: base: components_open: found loaded component ofi
[f011:125594] mca: base: components_open: component ofi open function successful
[f011:125594] mca:base:select: Auto-selecting mtl components
[f011:125594] mca:base:select:(  mtl) Querying component [psm]
[f011:125594] mca:base:select:(  mtl) Query of component [psm] set priority to 100
[f011:125594] mca:base:select:(  mtl) Querying component [ofi]
[f011:125594] mca:base:select:(  mtl) Query of component [ofi] set priority to 10
[f011:125594] mca:base:select:(  mtl) Selected component [psm]
[f011:125594] mca: base: close: component ofi closed
[f011:125594] mca: base: close: unloading component ofi
[f011:125594] select: initializing mtl component psm
```

Am I missing anything?

hppritcha:
@yburette 
what is your mpirun command line look like?

yburette:
mpirun -host f010,f011 -n 2 ./test

hppritcha:
please change to

mpirun -host f010,f011 -n 2 --mca mtl ofi ./test

hppritcha:
i'm assuming your working with the 1.10 release candidate.

yburette:
I'm working with this PR.

When I add --mca mtl ofi (but not --mca pml cm), I get the following error:

```
WARNING: There are more than one active ports on host 'f011', but the
default subnet GID prefix was detected on more than one of these
ports.  If these ports are connected to different physical IB
networks, this configuration will fail in Open MPI.  This version of
Open MPI requires that every physically separate IB subnet that is
used between connected MPI processes must have different subnet ID
values.

Please see this FAQ entry for more details:

  http://www.open-mpi.org/faq/?category=openfabrics#ofa-default-subnet-gid

NOTE: You can turn off this warning by setting the MCA parameter
      btl_openib_warn_default_gid_prefix to 0.
mpirun noticed that process rank 0 with PID 0 on node f011 exited on signal 11 (Segmentation fault).
```

Is this Seg fault related to the problem with inlining you mention in open-mpi/ompi/pull/747?

yburette:
Finally read the mailing list :-)

I added the call to opal_progress_unregister() in the OFI MTL's close() function and I don't see the Seg fault anymore:

``` diff
@@ -98,7 +98,18 @@ ompi_mtl_ofi_component_query(mca_base_module_t **module, int *priority)
 static int
 ompi_mtl_ofi_component_close(void)
 {
-    return OMPI_SUCCESS;
+    int ret;
+
+    /**
+     * Unregister progress callback.
+     */
+    ret = opal_progress_unregister(ompi_mtl_ofi_progress);
+    if (OMPI_SUCCESS != ret) {
+        opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
+                            "%s:%d: opal_progress_unregister failed: %d\n",
+                            __FILE__, __LINE__, ret);
+    }
+    return ret;
 }
```

PHHargrove:
@yburette 
Is the patch in the previous comment in git anywhere?
I need something I can script patching from, and the issue tracker doesn't cut it.

hppritcha:
@yburette the segfault you are reporting above appears to be something new.

jsquyres:
@yburette Does this address the potential inlining of the progress function (Howard mentioned he saw problems w.r.t. the inlining -- but it may have been related to not unregistering during close...?).

yburette:
@PHHargrove yes, I have just pushed it to my fork

yburette/ompi-release@4dc4d15a7f61b5251c7b49dbcad56dd459454caf

Hope this helps

yburette:
@hppritcha @jsquyres 
Ralph mentioned on the mailing list that not un-registering the progress callback could create a Seg Fault during Finalize.

This definitely seems to fix the issue I reported this morning.

Howard, what would be the best way for me to reproduce the issue w.r.t. inlining?

hppritcha:
yburette/ompi-release@4dc4d15 is equivalent to open-mpi/ompi#747  although the possible problem of passing an inline function to opal_progress_register may reoccur since the progress function is still marked as inlineable. 

I've never observed the "There are more than one active ports on host..." message.  I think that's something different, probably owing to the ofi/psm provider as well as the openib btl  trying to initialize.

PHHargrove:
@hppritcha @yburette 

Short version: Yohann's commit 4dc4d15 and open-mpi/ompi#747 are the same for me.

Longer version:

I previously reported on the mailing list this PR + open-mpi/ompi#747 worked when passing no args other than "-np 2" to mpirun.
As Howard observes, Yohann's commit 4dc4d15 is a subset of the changes in ompi#474.
So it should come as no surprise that this PR + yburette/ompi-release@4dc4d15 works just fine with no extra args.

However, I still see (just as reported on the mailing list with respect to 747) a SEGV if I run "mpirun -mca btl sm,self -np 1 examples/ring_c" unless I also add "-mca mtl ^ofi" or "-mca mtl portals4".

PHHargrove:
@yburette 

The "There are more than one active ports on host..." message is something I've seen before on systems with multiple ports on the same IB network, and is the subject of an [FAQ Entry](http://icl.cs.utk.edu/open-mpi/faq/?category=openfabrics#ofa-default-subnet-gid).  On systems I use that are connected this way, I set the MCA parameter described in that FAQ entry to disable the warning.

tkordenbrock:
@hppritcha 

#409 applied, built and tested OK with mtl-portals4 in a single MTL build.  All my setups are mtl-portals4 only, so I haven't been able to check if it segfaults if it is not selected.

I don't think mtl-portals4 is susceptible because progress is not registered until add_procs() which gets called after it is selected.  I'll put together a setup with multiple MTLs to confirm.

tkordenbrock:
@hppritcha 

#409 applied, built and tested OK with mtl-portals4 and mtl-ofi.  I used this command:

`$OMPI_110/bin/mpirun -n 2 -N 1 --mca pml cm --mca mtl_ofi_priority 100 --mca mtl_portals4_priority 50 ./sendrecv`

hppritcha:
@tkordenbrock 
thanks
@rhc54  I think this can be merged in now.

rhc54:
ok - thx!!

PR #413, opened 2015-07-23
v1.10: VERSION: proposed new .so version numbers

jsquyres:
This commit reflects an update in the Libtool version numbers for many of the major shared libraries that Open MPI produces.  These version numbers reflect the first in the v1.10.x series, and are designed to _not_ be backwards compatible with the v1.8.x series.  As such, most of the version numbers simply went from x:y:z to (x+5):0:0.  This accomplishes two things:
1. Makes the new version be non-backwards-compatible (in an ABI sense)
2. Leaves a "gap" in the "c"urrent value for future 1.8.x releases (if we need them)

(this commit currently contains superflous comments so that a reviewer can examine the changes easily -- these comments will be removed before the PR is merged to v1.10)

Fixes open-mpi/ompi#734

@goodell please review

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -82,16 +82,16 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=7:0:6
+libmpi_so_version=12:0:0
 libmpi_cxx_so_version=2:3:1
-libmpi_mpifh_so_version=7:0:5
+libmpi_mpifh_so_version=12:0:0
 libmpi_usempi_tkr_so_version=5:0:4
-libmpi_usempi_ignore_tkr_so_version=1:0:1
-libmpi_usempif08_so_version=6:0:6
-libopen_rte_so_version=7:5:0
-libopen_pal_so_version=8:1:2
+libmpi_usempi_ignore_tkr_so_version=6:0:0
+libmpi_usempif08_so_version=11:0:0
+libopen_rte_so_version=12:0:0
+libopen_pal_so_version=13:0:0
 libmpi_java_so_version=3:0:2
-liboshmem_so_version=3:1:0
+liboshmem_so_version=8:0:0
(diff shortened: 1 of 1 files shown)


VERSION
```diff
@@ -82,16 +82,28 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=7:0:6
+# JMS Version changes since 185fcfb71a9b933483c8766c285ec4aab87632f7
+# right after 1.8.6 went to rc1 (Mon May 25 07:17:04 2015 -0700) up to
+# and including ed11ec4446c817fa4a3151076b42d9df99356057 (Tue Jul 21
+# 11:03:38 2015 -0700).
```
goodell: I don't understand this comment.

VERSION
```diff
@@ -82,16 +82,28 @@ date="Unreleased developer copy"
...
+
+libmpi_so_version=12:0:0
+# libmpi changed: OLD 7:0:6
```
goodell: Not sure that "OLD" will make any sense to someone reading this file in 6 months.  Do we really want these comments in this file or should we just rely on Git in order to check this in the future?  If we do keep them, we need to tweak them so that people making future modifications know when/how to modify/delete these comments.

VERSION
jsquyres: This is just me recording the range in the git logs that I looked at to make these changes.

VERSION
jsquyres: Per the description of this PR, these comments (including the JMS comment) will be removed before the final merge.  There's really only intended here to help with the review -- e.g., provide rationale.

jsquyres:
@goodell and I chatted on the phone.  He's good with the ((c+=5), r=a=0) scheme.  I removed the superfluous explanatory comments, and we're now good to go.

:+1:

PR #416, opened 2015-07-23
Update extensions section to include cuda

rolfv:
Update MPI extensions section to include CUDA piece

(cherry picked from commit ompi-mpi/ompi@d184f260adddf9fab307adc94f2c8e08853485cf)

### DIFF
--- a/README
+++ b/README
@@ -709,6 +709,9 @@ Open MPI Extensions
   - cr: Provides routines to access to checkpoint restart routines.
     See ompi/mpiext/cr/mpiext_cr_c.h for a listing of availble
     functions.
+  - cuda: When the library is compiled with CUDA-aware support, it provides
+    two things.  First, a macro MPIX_CUDA_AWARE_SUPPORT. Secondly, the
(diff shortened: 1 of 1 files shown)


rolfv:
bot:milestone:v2.0.0
bot:assign: @jsquyres

jsquyres:
@rolfv Do you have a man page for your cuda extension?  If so, do you want to cite it in the README?  E.g., "For more information, see ....".

rolfv:
There is one.  Maybe we could just go with what we have now and then I will update it later.  I think I still want to make some other changes in the CUDA extensions like fix the return value and have them build even when CUDA-aware is not there but just have them set to the right values.

jsquyres:
@rolfv Sure, sounds fine -- just don't forget (but we're not going to immediately release v2.0.0, so you've got some time).

Can you do me a favor and file an issue against 2.0.0 with what you plan to do for 2.0.0 (i.e., the stuff you mentioned in your comment), so that we can track it?  Thanks.

PR #422, opened 2015-07-24
 Fix MPI_Finalize

rhc54:
This has a horrible history associated with it. Apparently, we committed a pull request that would have included a change to ompi_mpi_finalize that fixed an ordering problem - we were incorrectly finalizing the PML before finalizing the communicator system, and the latter causes PSM to send out messages. However, the change was immediately reverted because it caused the USNIC BTL to segfault.

Subsequently, the problem was fixed in the master. Sadly, the completed fix was never brought back to the 1.8 series.

We also found another MPI_Finalize ordering fix that came in later, but was not brought to the 1.8 series. We are including it here just in case it is needed.

The two commits in question:
(cherry picked from commit open-mpi/ompi@faf008f5277a2cb081010f6d284c5d1b9651b4d1)
(cherry picked from commit open-mpi/ompi@dee243c58d0c79faa445b3c12685fc02def27085)

@hjelmn @bosilca Please see if this is okay, and if we missed other things that need to be here.

### DIFF
--- a/ompi/mca/bml/r2/bml_r2.c
+++ b/ompi/mca/bml/r2/bml_r2.c
@@ -446,7 +446,11 @@ static int mca_bml_r2_del_procs(size_t nprocs,
 
     for(p = 0; p < nprocs; p++) { 
         ompi_proc_t *proc = procs[p]; 
-        if(((opal_object_t*)proc)->obj_reference_count == 1) { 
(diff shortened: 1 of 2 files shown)


rhc54:
Obviously, we are still missing something(s). :-(

I would appreciate help in tracking down the missing elements.

hjelmn:
From what I remember this is the correct fix. :+1:

PR #425, opened 2015-07-28
Merge pull request #745 from edgargabriel/pr/sharedfp-sm-logic3

edgargabriel:
@jsquyres since you reviewed the original pull request to master, maybe its easiest if you can have a look whether the pull request to v2.x is also complete and correct.


jsquyres:
FWIW, it would have been ok to leave this as the multiple commits that it originated as -- each commit was a small, individual thing (e.g., one commit was just about fixing the whitespace stuff, and was separate from all other code/logic changes).

:+1:

PR #426, opened 2015-07-28
Pr/ompio multiple smaller things

edgargabriel:


ompi/mca/io/ompio/io_ompio_file_open.c
```diff
@@ -379,8 +394,9 @@ int mca_io_ompio_file_delete (char *filename,
 
     ret = unlink(filename);
 
-    if (0 > ret) {
-        return OMPI_ERROR;
+    if (0 > ret && ENOENT != errno ) {
+	opal_output (1, "errno = %d %s\n", errno, strerror(errno));
+        return MPI_ERR_ACCESS;
```
hppritcha, line 399: could this be OMPI_MPI_ERR_ACCESS?  I'm not sure what the Open MPI convention is about what level returns an MPI error code vs an OMPI internal error return value.

edgargabriel:
you are probably right, will double check. The utilization of error codes is probably not as clean in the code base as it could be.

hppritcha:
Not critical about the error codes.  Just curious.

This passes for me as long as I avoid lustre.
:+1: 

@jsquyres modulo the error return codes, I think this is good to go.

PR #430, opened 2015-07-28
oob/tcp: fix a race condition when finalizing the oob/tcp component

ggouaillardet:
(cherry picked from commit open-mpi/ompi@429bdf1af7dd28b37237e5fd0fd61a2907a3c268)

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp_component.c
+++ b/orte/mca/oob/tcp/oob_tcp_component.c
@@ -16,6 +16,8 @@
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2013-2015 Intel, Inc.  All rights reserved.
  * Copyright (c) 2014      NVIDIA Corporation.  All rights reserved.
+ * Copyright (c) 2015      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -68,6 +70,7 @@
 #include "orte/util/parse_options.h"
 #include "orte/util/show_help.h"
 #include "orte/runtime/orte_globals.h"
+#include "orte/runtime/orte_wait.h"
 
 #include "orte/mca/oob/tcp/oob_tcp.h"
 #include "orte/mca/oob/tcp/oob_tcp_component.h"
@@ -630,10 +633,22 @@ static int component_startup(void)
     return rc;
 }
 
+static void cleanup(int sd, short args, void *cbdata)
+{
+    opal_list_item_t * item;
+    bool *active = (bool*)cbdata;
+    while (NULL != (item = opal_list_remove_first(&mca_oob_tcp_component.listeners))) {
+        OBJ_RELEASE(item);
+    }
+    if (NULL != active) {
+        *active = false;
+    }
+}
+
 static void component_shutdown(void)
 {
     int i = 0;
-    opal_list_item_t *item;
+    bool active;
 
     opal_output_verbose(2, orte_oob_base_framework.framework_output,
                         "%s TCP SHUTDOWN",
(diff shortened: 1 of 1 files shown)


ggouaillardet:
:bot:assign: @rhc54 
:bot:milestone:v2.0.0
:bot:label:bug

hppritcha:
It appears this commit is actually associated with a bunch of other commits in to master that haven't been put in to v2.x: open-mpi/ompi@61fb067f1, open-mpi/ompi@683efcb85, and open-mpi/ompi@219c4dfba

Unfortunately, having this commit in v2.x without the others - that I guess this commit was trying to fix, breaks things.

For launches with, for example, aprun on cray's,  v2.x branch currently has a failure to land properly -hangs running a hello world with 1 rank.  With the missing commits in v2.x, things work again.

jsquyres:
Just to close the loop: @hppritcha filed #457 to fix this.

rhc54:
I approved this only because (a) it is technically correct, and (b) Gilles seemed to think it necessary. I personally would not recommend that #457 go into the 2.x branch, and this one should probably be reverted as well. There are still known issues on the master due to the async/sync thread split.

ggouaillardet:
without this PR, and on OS X (Yosemite, 10.10.4) i can get such warnings with a dumb hello world program

```
# mpirun -np 1 ./hw
[warn] select: Bad file descriptor
I am 0 / 1
```

the root cause is the main thread closes a file descriptor while the async thread is running select with this file descriptor. (and select returning with an error in such a case could be OS X specific)

in master, there is only one async thread, but in v2.x, there are two (orte and pmix_native)
in this case, i talk about the orte async thread.

here is the relevant part of code for this PR

``` c
    /* because the listeners are in a separate
     * async thread for apps, we can't just release them here.
     * Instead, we push it into that event thread and release 
     * them there */
     if (ORTE_PROC_IS_APP) {
        opal_event_t ev;
        active = true;
        opal_event_set(orte_event_base, &ev, -1,
                       OPAL_EV_WRITE, cleanup, &active);
        opal_event_set_priority(&ev, ORTE_ERROR_PRI);
        opal_event_active(&ev, OPAL_EV_WRITE, 1);
        ORTE_WAIT_FOR_COMPLETION(active);
     } else {
        /* we can call the destruct directly */
        cleanup(0, 0, NULL);
     }
```

i do not know what happens with aprun ...
no orte thread at all ? orte thread already finished ?
in both cases, that can explain the hang ...
i guess the right fix is to replace `ORTE_PROC_IS_APP` by something like
`ORTE_PROC_IS_APP && orte_thread_is_up_and_running()`

now if there is a sleeping orte thread, is something missing to wake it up ?

rhc54:
I think the right thing is to correctly fix the master, and then figure out what needs to be brought across to the release branch. I don't know what was already incorrectly brought across, but best to leave it until we have things right.

hppritcha:
im going to reverr this pr.  it breaks cray and lanl needs this branch to work for trinity testing.

orte/mca/oob/tcp/oob_tcp_component.c
```diff
@@ -644,16 +659,37 @@ static void component_shutdown(void)
...
     }
 
-    while (NULL != (item = opal_list_remove_first(&mca_oob_tcp_component.listeners))) {
-        OBJ_RELEASE(item);
-    }
+    /* because the listeners are in a separate
+     * async thread for apps, we can't just release them here.
+     * Instead, we push it into that event thread and release 
+     * them there */
+     if (ORTE_PROC_IS_APP) {
```
ggouaillardet, line 671: @rhc54 should it be
if (!ORTE_PROC_IS_HNP)
instead ?

hppritcha:
reverting

orte/mca/oob/tcp/oob_tcp_component.c
ggouaillardet, line 671: @rhc54 let me take it back, the line could be

```
if (ORTE_PROC_IS_APP && orte_initialized)
```

i can reproduce the hang without mpirun (e.g. singleton).
behavior of master and v2.x are differents, and the new test is valid for v2.x only.

if mpirun -np 1 ./a.out
then component_shutdown is invoked at orte_init() time (i really meant init and not finalize)
on both master and v2.x, the async progress is alive and kicking
now, if ./a.out (singleton), then component_shutdown is invoked at orte_finalize() time.
progress_thread is alive on master, but has already finished on v2.x (and hence the hang)

bottom line, replacing

```
if (ORTE_PROC_IS_APP)
```

with

```
if (ORTE_PROC_IS_APP && orte_initialized)
```

works on v2.x only.
The use of `orte_initialized` is everything but intuitive here (at least for me) but i could not find a better variable.
i was able to test this on a cray with aprun, and it also removed the hang.

this PR removes an annoying warning i could only observe on OSX.
i can make a new one-off PR for v2.x or wait until the master is stabilized and related commits land into the v2.x branch.

any thoughts ?

rhc54:
I'd just let it rest for now. I'll take a look at master here shortly, and then figure out how to bring things across cleanly to 2.0. For now, this isn't an issue given that we aren't close to a 2.0 release anyway.

PR #433, opened 2015-07-29
Topic/v1.10/f08 bindings and makefiles

ggouaillardet:
:bot:assign: @jsquyres
:bot:milestone:v1.8.8
:bot:label:bug


ggouaillardet:
:bot:milestone:v1.10.0

jsquyres:
:+1:

jsquyres:
The more I think about this, the less comfortable I am with it -- only because we have a potentially very short release cycle for v1.10.  The code in this PR is "new" -- it hasn't soaked for a long time out in the real world.

To be safe / conservative, we should revert this PR and then apply the same fix for v1.10 that we did for v1.8 (i.e., #437).

@rhc54 Ok to revert this PR?

rhc54:
I defer to your judgement on all things Fortran - feel free to revert and fallback to the 1.8 patch

PR #434, opened 2015-07-29
Topic/v2.x/f08 bindings and makefiles

ggouaillardet:
:bot:assign: @jsquyres
:bot:milestone:v2.0.0
:bot:label:bug

PR #437, opened 2015-07-30
fortran: update sentinels-checking macros

jsquyres:
There is a lengthy comment in the code explaining what this change is and why it is being done.  It is a one-off for the v1.8.8 release (i.e., we do stuff different in >=v1.10.0, so this change isn't relevant anywhere else).

@ggouaillardet I'm proposing this PR as the v1.8.x-specific fix for https://github.com/open-mpi/ompi/issues/739 instead of your PR https://github.com/open-mpi/ompi-release/pull/434.  This fix is much smaller / safer (since we want to release v1.8.8 in the immediate future), and it also doesn't break ABI.

I have verified that this fix makes MPI_STATUSES_IGNORE (and friends) be correctly identified when there are no weak symbols.

@ggouaillardet please review (in the immediate future, please! :smile: )

### DIFF
--- a/ompi/mpi/fortran/base/constants.h
+++ b/ompi/mpi/fortran/base/constants.h
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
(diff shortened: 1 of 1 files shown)


ggouaillardet:
:+1:

PR #438, opened 2015-07-31
Fix an invalid memory access in mrecv & imrecv

rhc54:
After the call to PML mrecv/imrecv, the message handle
is set to MPI_MESSAGE_NULL. Use the cached communicator
while invoking the error handler.

(cherry picked from commit open-mpi/ompi@5a7bd898f9c53876063e60a6ee2734c85da7dc68)


hppritcha:
@rhc54 @jsquyres looks like this also needs to be applied to v2.x

PR #446, opened 2015-07-31
Fix an invalid memory access in mrecv & imrecv

hppritcha:
After the call to PML mrecv/imrecv, the message handle
is set to MPI_MESSAGE_NULL. Use the cached communicator
while invoking the error handler.

@jsquyres 

cherrry -pick of open-mpi/ompi@5a7bd89


jsquyres:
:+1:

PR #452, opened 2015-08-01
Fix an annoying segfault caused by incorrect indentation

rhc54:
Fix an annoying segfault caused by incorrect indentation in a loop that causes the buffer to not be created prior to packing.

(cherry picked from commit open-mpi/ompi@9bc384282a310421c08df54bbc6a3d1307d014ce)

@jsquyres This should fix the no-disconnect problem. It is actually a fairly small change, but caused lots of re-indentation.

### DIFF
--- a/orte/orted/pmix/pmix_server_process_msgs.c
+++ b/orte/orted/pmix/pmix_server_process_msgs.c
@@ -330,8 +330,8 @@ void pmix_server_process_message(pmix_server_peer_t *peer)
          * was actually provided so we don't hang if no modex data is being given */
         OPAL_LIST_FOREACH_SAFE(req, nextreq, &pmix_server_pending_dmx_reqs, pmix_server_dmx_req_t) {
             if (0 == opal_compare_proc(id, req->target)) {
-                /* yes - deliver a copy */
                 reply = OBJ_NEW(opal_buffer_t);
+                /* yes - deliver a copy */
                 if (NULL == req->proxy) {
                     /* pack the status */
                     ret = OPAL_SUCCESS;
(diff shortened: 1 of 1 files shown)


hppritcha:
Was this caused by the white space tool?  I'm confused about the title of the PR.

rhc54:
No, it was caused by an obj_new being in the wrong level of indentation.

jsquyres:
The no-disconnect test still hangs, but now when you kill mpirun, the procs die, too.

Thanks!

:+1:

rhc54:
Really? It no longer hangs on master for me, so it looks like something else must also be missing on 2.x. What I changed here should in no way have impacted the "suicide" operation - it was intended to solve the no-disconnect hang, which it fixed for me.

jsquyres:
Let me try the test on master...

jsquyres:
Ah, I think I see the problem.

I'm running under SLURM, and if I don't specify `--oversubscribe` on the `mpirun` command line, I get this error in the middle of the test:

```
All nodes which are allocated for this job are already filled.
```

and the test hangs there (assumedly because MPI_COMM_SPAWN is trying to spawn, but ORTE says "Nay nay! There art too many processes on thine nodes!"... but then doesn't actually return a failure up to MPI_COMM_SPAWN, which would trigger killing the jobs.  (this is a guess)

If I add `--oversubscribe` to the `mpirun` command line, then the test completes as normal.

So I think this PR is good for fixing the first problem.  Should we have a different issue/PR for the hanging issue?

hppritcha:
@jsquyres are you good with merging this?

rhc54:
@jsquyres Yeah, that's a separate issue that needs to be addressed. Nothing to do with this one.

jsquyres:
:+1:

PR #458, opened 2015-08-03
mtl/ofi: Merge some changes from master.

yburette:


jsquyres:
Something didn't go right (I don't have time to investigate at the moment):

```
make[2]: Entering directory `/home/jsquyres/git/ompi-crossover/ompi/mca/mtl/ofi'
make[2]: *** No rule to make target `/opal/mca/common/libfabric/lib@OPAL_LIB_PREFIX@mca_common_libfabric.la', needed by `mca_mtl_ofi.la'.  Stop.
make[2]: *** Waiting for unfinished jobs....
```

yburette:
I removed the common libfabric commit until I figure out what was missing...

hppritcha:
I'm installing libfabric on the IU jenkins master node so we can catch problems like this with the disable-dlopen check.

jsquyres:
@yburette Perhaps it isn't worth it for the libfabric common commit on the v1.10 branch (I know I said the opposite, yesterday).  But a) since it's going to involve at least some level of porting, and b) the v1.10 series isn't expected to have a hugely long life, perhaps the Way Forward is 2.x -- and that's where we can have the libfabric common stuff.

Howzat sound?

yburette:
@jsquyres Fine by me. I've already removed the common libfabric commit from this PR.
I was able to build and test it with the PSM provider.

jsquyres:
:+1:

PR #459, opened 2015-08-04
Revert "oob/tcp: fix a race condition when finalizing the oob/tcp com…

hppritcha:
Reverts open-mpi/ompi-release#430


hppritcha:
per discussion at today's dev meeting, reverting the oob/tcp PR #430.  A more comprehensive solution will be implemented on master and pushed back to the 2.x branch.

PR #16, opened 2014-10-09
Adding a new OPAL hash table routine. Please read the algorithm descript...

jladd-mlnx:
...ion in opal/class/opal_hash_table.c for more precise details on the design and implementation. This algorithm was contributed by David Linden of H.P. in partnership with Mellanox Technologies. This contribution achieves two objectives:
1. It's actually hashing now, whereas the old OPAL hash table was not. Thus, it is a bug fix and, as such, should be included in the 1.8 series.
2. It is dynamic and can grow and shrink the number of buckets in accordance with job size, whereas the old OPAL hash table had a fixed number of buckets which resulted in poor retrieval performance at large scale.

This scheme has been deployed in the field on very large H.P./Mellanox systems and has been demonstrated to significantly decrease job start-up time (~ 20% improvement) when launching applications directly with srun in SLURM environments. However, neither SLURM nor direct launch are prerequisites to take advantage of this change as any entity that utilizes OPAL hash table objects can benefit (at least partially) from this contribution.

PR #14, opened 2014-10-07
Fix the application abort routine so we actually abort

jladd-mlnx:
Adding @bureddy 's solution for OSHMEM's global exit when applications are launched directly with srun.

I don't think this is the final solution, but I don't want this to get lost in the shuffle. @jsquyres @miked-mellanox @rhc54 

CMR Trac ticket:
https://svn.open-mpi.org/trac/ompi/ticket/4919

PR #463, opened 2015-08-04
mtl/ofi: Merge inline progress function from master

yburette:


hppritcha:
double checking that this works for me on our infinipath cluster here.

hppritcha:
This works for me for the sockets provider.  I'm using the ring_c.c test in examples and using libfabric at
ofiwg/libfabric@465d02f.  When using the psm provider, I get a segfault in mpi_finalize once ever few runs using this command line:

mpirun -np 4 -N 4 --mca pml cm --mca mtl ofi  --mca mtl_ofi_provider psm ./ring_c

The BT for the second thread (first thread caught the signal and is trying to print out a backtrace) looks like:

#0  0x00002b0b71062bd0 in _dl_close_worker (map=<value optimized out>) at dl-close.c:149
#1  0x00002b0b710637fe in _dl_close (_map=0xf80650) at dl-close.c:757
#2  0x00002b0b7105d266 in _dl_catch_error (objname=0xef3760, errstring=0xef3768, mallocedp=0xef3758, operate=0x2b0b72600fe0 <dlclose_doit>, args=0xf80650) at dl-error.c:178
#3  0x00002b0b7260129c in _dlerror_run (operate=0x2b0b72600fe0 <dlclose_doit>, args=0xf80650) at dlerror.c:164
#4  0x00002b0b7260100f in __dlclose (handle=<value optimized out>) at dlclose.c:48
#5  0x00002b0b71f3b62b in dlopen_close (handle=0xf80ba0) at dl_dlopen_module.c:141
#6  0x00002b0b71f3b01c in opal_dl_close (handle=0xf80ba0) at base/dl_base_fns.c:53
#7  0x00002b0b71f132b8 in mca_base_component_repository_release_internal (ri=0xedf840) at mca_base_component_repository.c:277
#8  0x00002b0b71f13346 in mca_base_component_repository_release (component=0x2b0b7929f460) at mca_base_component_repository.c:300
#9  0x00002b0b71f1471f in mca_base_component_unload (component=0x2b0b7929f460, output_id=-1) at mca_base_components_close.c:46
#10 0x00002b0b71f14784 in mca_base_component_close (component=0x2b0b7929f460, output_id=-1) at mca_base_components_close.c:59
#11 0x00002b0b71f1480c in mca_base_components_close (output_id=-1, components=0x2b0b715cd5b0, skip=0x0) at mca_base_components_close.c:85
#12 0x00002b0b71f147b3 in mca_base_framework_components_close (framework=0x2b0b715cd560, skip=0x0) at mca_base_components_close.c:65
#13 0x00002b0b713634c0 in mca_pml_base_close () at base/pml_base_frame.c:162
#14 0x00002b0b71f226a4 in mca_base_framework_close (framework=0x2b0b715cd560) at mca_base_framework.c:208
#15 0x00002b0b712c2c1e in ompi_mpi_finalize () at runtime/ompi_mpi_finalize.c:316
#16 0x00002b0b712ec0d1 in PMPI_Finalize () at pfinalize.c:46

Only way I've found to work around the problem when using the psm provider is to configure with --disable-dlopen.

hppritcha:
:question:  

I'm okay with merging this as the problem with the segfaults appears to be related to the PSM provider.
I also tried hacking out the entire mtl/psm provider to avoid use of psm (even via the dlopen of the psm component), but the segfault still happens.

hppritcha:
@jsquyres I'd say lets go ahead and merge this.  There's something broken with the psm provider, at least for me.

Here's the version info from the psm.h header file:

/\* MQ handle (opaque)
 *
 *
- Handle returned to the user when a new Matched queue is created (@ref
- psm_mq_init).  */
  typedef struct psm_mq *psm_mq_t;

#define PSM_VERNO       0x010e
#define PSM_VERNO_MAJOR 0x01
#define PSM_VERNO_MINOR 0x0e

jsquyres:
@hppritcha I'm not sure I understand.  You're saying that there's still a segv, right?  Does this segv only occur in certain cases?  Or will it always occur if libfabric is compiled against PSM _at all_?

hppritcha:
I only see the segv if I

1) request cm pml
2) request mtl ofi
3) don't specifiy provider with mca parameter (in which case psm is the
first in the list and gets picked by the ofi mtl)

I always build libfabric with the verbs (which doesn't count here since it
doesn't support FI_EP_RDM), sockets, and psm providers.

If I select the sockets provider for the ofi mtl (via the ofi mca param),
the segfault never appears.  So its something going on with the PSM provider
within libfabric that seems to be the root cause.

2015-08-06 11:04 GMT-06:00 Jeff Squyres notifications@github.com:

jsquyres:
@hppritcha What version of libfabric are you using?

Intel: @yburette et al., I thought the point of the OFI MTL was to be able to use PSM, right?  If so, it would somewhat defeat the point to merge this PR in if it breaks PSM.  Right?

hppritcha:
libfabric top of master (or at ofiwg/libfabric@465d02f) as of this morning.

The system I'm using has psm version -
#define PSM_VERNO 0x010e
#define PSM_VERNO_MAJOR 0x01
#define PSM_VERNO_MINOR 0x0e

I looked for obvious suspects like passing pointers to psm provider internal strings back in `fi_info` but nothing obvious stuck out.

jsquyres:
@yburette Any insight here?

yburette:
Why do I see this only now, I don't know :-(

Let me look into it, I thought I had verified it with the PSM provider before creating this PR.

yburette:
I just checked again with latest ompi-release/v2.x and libfabric. Both sockets and PSM providers work for me.

hppritcha:
what version of psm are you using?  It might be that I need to have a newer psm installed on our cluster.

yburette:
Hopefully, that's the problem.

I'm using the one that comes with OFED-3.18:
infinipath-psm-3.3-2_g6f42cdb_open

hppritcha:
Hmm... interesting.

The DOE TOSS systems do their own packaging of things, but looks like on
the systems I'm using we have
infinipath-psm-3.1c-3chaos...

Could I get the 3.3-2 rpm and install psm library in an alternate location
and expect it to work with the
kernel part of whatever comes with infinipath-psm-3.3-2?

2015-08-12 9:13 GMT-06:00 Yohann Burette notifications@github.com:

yburette:
Good question... Let me ask internally and I'll get back to you.

In the meantime, you can always try it :-)

yburette:
From what I understand, there should be no problem with installing a newer version of PSM in a different location and setting LD_LIBRARY_PATH accordingly.

yburette:
I also assume that libfabric and the PSM provider work outside of OpenMPI, right?

hppritcha:
I didn't check all the fabtests, but the ones I checked with ep rdm worked for me with psm provider.

hppritcha:
I switched to using the libpsm_inifinipath from OFED 3.1.8.  I still observe the same issue - but not with every run - where I observe a segfault in MPI_Finalize unless I configure with --disable-dlopen.  Again it seems to be something specific to the PSM provider.

Since @yburette cannot reproduce this, I say we go ahead and merge.

yburette:
@hppritcha We are still trying to reproduce the problem on our machines.

Does the segfault in MPI_Finalize() with the PSM provider go away when using the following?

-x FI_PSM_NAME_SERVER=0

hppritcha:
bingo, that fixes the problem!

2015-08-13 10:13 GMT-06:00 Yohann Burette notifications@github.com:

yburette:
@hppritcha Great! Yes, it's a current problem with the PSM provider. Hopefully, there'll be a fix for it soon. Note that it must be some sort of race condition because I can't reproduce the issue on our setup.

hppritcha:
@jsquyres this can be merged.

PR #473, opened 2015-08-06
CSCuv67889: usnic: fix an error corner case

jsquyres:
Ensure that we have non-NULL on all levels of pointers, which will save us if there are exitable errors very early during component / module initialization.

(cherry picked from commit open-mpi/ompi@b5c37dbfe2ce6353c756d161c1cd4e75d5f5f127)

@bturrubiates please review


bturrubiates:
:+1:

bturrubiates:
:+1:

PR #482, opened 2015-08-10
usnic: sync with master

jsquyres:
Sync with the latest master usnic BTL, including the OPAL async progress thread, the new opal_sync_event_base name, and fixing a race between the connectivity checker agent shutting down before all clients could connect.


rhc54:
:+1:

jsquyres:
@rhc54 and I chatted about this -- the code looks good, but I want to check last night's MTT to ensure that the introduction of this code on master didn't cause a bunch of failures.

rhc54:
I took off the RM-approved tag to remind me that we don't want this committed to 1.10 until Jeff clears the last night's MTT.

jsquyres:
The MTT failures were caused by a bad option in an MTT INI file -- not this new code.  So we're clear to merge this PR.

PR #484, opened 2015-08-11
Make the async copy the default.

rolfv:
This is a one character change that makes the async copy the default.  This has been the default in the master version for almost a year so it is time to put it in the 1.10 series also.
@jsquyres Can you review?
@rhc54 I need this in 1.10 FYI


hppritcha:
does this also need to go in to v2.x?

rolfv:
This does not need to go to v2.x.  It is already there as well as in the master.

PR #498, opened 2015-08-13
Cleanup hetero ops so they don't segfault

rhc54:
no correlating change on master because things are very different there.


rhc54:
@jsquyres can you please take a look - trivial change

jsquyres:
:+1:

PR #503, opened 2015-08-14
btl/openib: remove OFED version runtime check when XRC is used

ggouaillardet:
this test seems broken :
- some false positive were reported
- it fails to detect some OFED version mismatch
  this commit simply removes this test, which means the application
  will likely fail if XRC is used ad OFED version is different
  between compile time and runtime

(cherry picked from commit open-mpi/ompi@d02ccd67de45145bfe0f858f80e092b152ba379d)

### DIFF
--- a/opal/mca/btl/openib/btl_openib_component.c
+++ b/opal/mca/btl/openib/btl_openib_component.c
@@ -1011,10 +1011,6 @@ static void device_destruct(mca_btl_openib_device_t *device)
 
 #if HAVE_XRC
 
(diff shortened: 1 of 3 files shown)


ggouaillardet:
:bot:assign: @rhc54 
:bot:label:bug
:bot:milestone:v1.10.0

ggouaillardet:
:bot:milestone:v2.0.0

rhc54:
@bureddy reviewed the identical fix for 1.10

hppritcha:
@jsquyres I think this is ready to merge

PR #507, opened 2015-08-14
orte oob ud: OPAL_LIB_PREFIX does not exist in v1.10

jsquyres:
@rhc54 please review


rhc54:
Hmmm...actually, the ud component is completely ompi_ignore'd. Perhaps we should just remove it entirely?

jsquyres:
@rhc54 Up to you. :smile:   I found the problem while looking for something else, so I thought I'd file this PR... Feel free to close / replace it with another that rm's the oob/ud component from v1.10...

rhc54:
Okay, I'll do that.

PR #508, opened 2015-08-14
Makefile.am's: if calling OPAL functions, must link to it

jsquyres:
On some OSs (e.g., Ubuntu 14.04.2 LTS), the linker is configured such that the symbols of library dependencies are not available to the application.  Hence, we need to explicitly list such dependencies when creating the executable.

For this commit, these tests are use OPAL function calls, so we must explicitly link in libopen-pal.so.

(cherry picked from commit open-mpi/ompi@42b9a966d622c190e65c2f3fbbc998da7b066f6f)

### DIFF
--- a/ompi/debuggers/Makefile.am
+++ b/ompi/debuggers/Makefile.am
@@ -43,7 +43,9 @@ headers = \
 # Simple checks to ensure that the DSOs are functional
 
 dlopen_test_SOURCES = dlopen_test.c
-dlopen_test_LDADD = $(top_builddir)/ompi/libmpi.la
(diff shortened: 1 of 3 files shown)


jsquyres:
This is similar in spirit to open-mpi/ompi#711

jsquyres:
@rhc54 please review (this is the v2.x version of #509)

rhc54:
:+1:

jsquyres:
Howard is out today, so I'll merge (especially since the LANL/IU Jenkins may not complete its build).

PR #509, opened 2015-08-14
Makefile.am's: if calling OPAL functions, must link to it

jsquyres:
On some OSs (e.g., Ubuntu 14.04.2 LTS), the linker is configured such that the symbols of library dependencies are not available to the application.  Hence, we need to explicitly list such dependencies when creating the executable.

For this commit, these tests are use OPAL function calls, so we must explicitly link in libopen-pal.so.

(ported from open-mpi/ompi@42b9a966d622c190e65c2f3fbbc998da7b066f6f)

@rhc54 please review

### DIFF
--- a/ompi/debuggers/Makefile.am
+++ b/ompi/debuggers/Makefile.am
@@ -43,7 +43,9 @@ headers = \
 # Simple checks to ensure that the DSOs are functional
 
 dlopen_test_SOURCES = dlopen_test.c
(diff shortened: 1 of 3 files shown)


rhc54:
:+1:

jsquyres:
IU / LANL is having some problems with their Jenkins; the LANL builds may or may not complete.  I think "default" is Mellanox; that one should still likely complete...?

PR #516, opened 2015-08-15
fortran: link the opal-pal.la library directly

jsquyres:
The `libmpi_*.la` fortran libraries make some direct calls to `libopen-pal.la`.  In many (most?) cases, having `libmpi_*` link against `libmpi` is sufficient (because `libmpi` pulls in `libopen-pal`). But when building RPMs on SLES, some compiler/linker flags are used that seem to make this implicit linking not sufficient -- we get missing opal symbols when creating `libmpi_mpifh.la`.  So link in `open-pal` directly (vs. indirectly), which solves the problem.

(cherry picked from commit open-mpi/ompi@a7f897ae49a336bcc51ce4af362553c8aabde04a)

@miked-mellanox Please review -- this is a PR for open-mpi/ompi#711

### DIFF
--- a/ompi/mpi/fortran/mpif-h/Makefile.am
+++ b/ompi/mpi/fortran/mpif-h/Makefile.am
@@ -9,7 +9,7 @@
 #                         University of Stuttgart.  All rights reserved.
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
-# Copyright (c) 2006-2014 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2006-2015 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2011-2013 Inria.  All rights reserved.
 # Copyright (c) 2011-2013 Universite Bordeaux 1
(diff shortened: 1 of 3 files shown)


jsquyres:
@ggouaillardet can you comment/review?  These are the two comments from master.

ggouaillardet:
@jsquyres how can i reproduce the initial issue ?
i tried rpmbuild on opensuse 13.2 (with the v2.x branch) but did not face any issue
iirc, the problem was also reported on ubuntu. and i also had no issue with configure/make/make install

jsquyres:
You'll have to ask @miked-mellanox -- I was never able to replicate the issue.  See open-mpi/ompi#610 (that PR started as an email from Mike to me; I turned it into a PR, which is why it looks like the conversation started in the middle).

hppritcha:
@jsquyres is this ready to merge?

jsquyres:
@hppritcha Waiting for @ggouaillardet's review

ggouaillardet:
:+1 I will try to investigate on sles11sp4 later

ggouaillardet:
:bot:label:reviewed

PR #528, opened 2015-08-20
Java GC bugfix

nrgraham23:
This pull request adds two Buffer references to the
Request class. Whenever a request object is created
that has associated buffers, the buffers should be added
to these references so the java garbage collector does
not dispose of the buffers prematurely.

This is a more robust expansion on the idea first proposed by
@ggouaillardet, and further discussed in open-mpi/ompi#369.

Fixes open-mpi/ompi#369

Manual cherry-pick from open-mpi/ompi@d363b5d8326c33b275684a2a097162fc63c0e62e

Signed-off-by: Nathaniel Graham ngraham@lanl.gov

### DIFF
--- a/ompi/mpi/java/java/Comm.java
+++ b/ompi/mpi/java/java/Comm.java
@@ -620,7 +620,9 @@ public final Request iSend(Buffer buf, int count,
 {
     MPI.check();
     assertDirectBuffer(buf);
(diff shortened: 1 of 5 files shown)


jsquyres:
:+1: 

@osvegis @hppritcha The goal here is to get the fix to the long-standing open-mpi/ompi#369 into the v1.10 release.

PR #529, opened 2015-08-20
fortran sizeof: ensure mpi_sizeof*f90 is not empty

jsquyres:
Per http://www.open-mpi.org/community/lists/devel/2015/08/17775.php, some compilers don't like it when there's a .f90 file that only contains comments (and no actual Fortran code).  So if OMPI determines that the Fortran compiler does not support enough Fortran mojo to support MPI_SIZEOF, generate at least one dummy Fortran subroutine that can be compiled in an otherwise barren Fortran landscape that is devoid of life and hope.

(cherry picked from commit open-mpi/ompi@d5763a8288c994e1d55a333d45f1a85d64341aff)

Thanks to @PHHargrove for raising the issue.

### DIFF
--- a/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
+++ b/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
@@ -227,16 +227,24 @@ sub output_file {
 ! compiler to build Open MPI.\n\n";
 
         if ($want_bodies) {
+            my $name = "ompi_sad_panda";
+            $name = "pompi_sad_panda"
+                if ($pmpi_arg);
             print OUT "!
 ! Dummy subroutine, just so that there is *some* Fortran in this file
-! (some compilers are unhappy if there are no Fortran statements in this
+! (this is defensive programming: since the Fortran compiler doesn't
+! support enough mojo, configure should set some AM_CONDITIONALs such
+! that this file should not end up being compiled, but just in case
+! that logic changes someday and this file *does* end up getting
+! compiled, make sure that it's not entirely empty because some
+! compilers are unhappy if there are no Fortran statements in this
 ! file).
-subroutine ompi_sad_panda()
+subroutine $name()
   implicit none
 
   print *, 'Open MPI is a sad panda because your Fortran compiler'
   print *, 'does not support enough Fortran mojo for MPI_SIZEOF'
-end subroutine ompi_sad_panda\n\n";
+end subroutine $name\n\n";
         }
     }
 
--- a/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
+++ b/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/env perl
 #
-# Copyright (c) 2014 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2014-2015 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2015      Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
@@ -218,12 +218,26 @@ sub output_file {
 ! Specifically: we need support for the INTERFACE keyword,
 ! ISO_FORTRAN_ENV, and the STORAGE_SIZE() intrinsic on all types.
 ! Apparently, this compiler does not support both of those things, so
-! this file will be blank (i.e., we didn't bother generating the
-! necessary stuff for MPI_SIZEOF because the compiler doesn't support
+! this file will be (effecitvely) blank (i.e., we didn't bother
+! generating the necessary stuff for MPI_SIZEOF because the compiler
+! doesn't support
 ! it).
 !
 ! If you want support for MPI_SIZEOF, please use a different Fortran
 ! compiler to build Open MPI.\n\n";
+
+        if ($want_bodies) {
+            print OUT "!
+! Dummy subroutine, just so that there is *some* Fortran in this file
+! (some compilers are unhappy if there are no Fortran statements in this
+! file).
+subroutine ompi_sad_panda()
+  implicit none
+
+  print *, 'Open MPI is a sad panda because your Fortran compiler'
+  print *, 'does not support enough Fortran mojo for MPI_SIZEOF'
+end subroutine ompi_sad_panda\n\n";
+        }
     }
 
     close(OUT);
--- a/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
+++ b/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
@@ -227,9 +227,7 @@ sub output_file {
 ! compiler to build Open MPI.\n\n";
 
         if ($want_bodies) {
-            my $name = "ompi_sad_panda";
-            $name = "pompi_sad_panda"
-                if ($pmpi_arg);
+            my $name =  $pmpi_arg ? "pompi_sad_panda" : "ompi_sad_panda";
             print OUT "!
 ! Dummy subroutine, just so that there is *some* Fortran in this file
 ! (this is defensive programming: since the Fortran compiler doesn't
--- a/ompi/mpi/fortran/mpif-h/Makefile.am
+++ b/ompi/mpi/fortran/mpif-h/Makefile.am
@@ -109,13 +109,13 @@ MOSTLYCLEANFILES = *.mod
 # out this workaround indefinitely.  So Jeff advised Absoft to fix
 # this bug in Sep 2014.
 if OMPI_BUILD_FORTRAN_MPIFH_BINDINGS
-noinst_LTLIBRARIES += libmpi_mpifh_sizeof.la
 
 if BUILD_FORTRAN_SIZEOF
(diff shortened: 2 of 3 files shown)


jsquyres:
Mmmph.  Somehow the "x" bit got lost on the script.  Fixed -- jenkins will retest shortly...

rhc54:
@jsquyres any suggestions on how this gets reviewed? Is Paul going to pull this down and test it?

PHHargrove:
@rhc54 Yes, Jeff gave me (on the devel list) instructions to pull the updated perl script and apply to the RC tarball.  I am testing now.

PHHargrove:
No joy. :-1: 

As posted on the devel list, there is still a bogus libtool command which is trying to construct a library with no objects as input.

PHHargrove:
To clarify my down-vote on this PR:

Because of the bogus libtool command, it is clear that the changes in this  PR are not _sufficient_ to resolve the original problem.  However, it is also uncertain (at least in my tiny mind) whether or not Jeff's changes are even _necessary_ since they only change the (generated) source for an object that is (still) not linked into any library.

PHHargrove:
@jsquyres

I tried a hand-edit to the generated Makefile:

``` diff
--- ompi/mpi/fortran/mpif-h/Makefile.orig       Thu Aug 20 20:06:33 2015
+++ ompi/mpi/fortran/mpif-h/Makefile    Thu Aug 20 19:49:53 2015
@@ -1276,7 +1276,7 @@
 am_libmpi_mpifh_la_rpath =  \
        -rpath $(libdir)
 libmpi_mpifh_sizeof_la_LIBADD =
-#nodist_libmpi_mpifh_sizeof_la_OBJECTS = sizeof_f.lo
+nodist_libmpi_mpifh_sizeof_la_OBJECTS = sizeof_f.lo
 libmpi_mpifh_sizeof_la_OBJECTS =  \
        $(nodist_libmpi_mpifh_sizeof_la_OBJECTS)
 am_libmpi_mpifh_sizeof_la_rpath =
```

The result was multiply-defined pandas:

```
/bin/sh ../../../../libtool  --tag=CC   --mode=link cc  -m32 -D_REENTRANT -g  -version-info 12:0:0  -o libmpi_mpifh.la -rpath /export/home/phargrov/OMPI/openmpi-1.10.0rc3-solaris11-x86-ss12u4/INST/lib   ../../../../ompi/libmpi.la  profile/libmpi_mpifh_pmpi.la libmpi_mpifh_sizeof.la -lm -lsocket -lnsl
libtool: link: cc -G -h libmpi_mpifh.so.12 -o .libs/libmpi_mpifh.so.12.0.0  -z allextract profile/.libs/libmpi_mpifh_pmpi.a ./.libs/libmpi_mpifh_sizeof.a -z defaultextract  -R/export/home/phargrov/OMPI/openmpi-1.10.0rc3-solaris11-x86-ss12u4/BLD/ompi/.libs -R/export/home/phargrov/OMPI/openmpi-1.10.0rc3-solaris11-x86-ss12u4/BLD/orte/.libs -R/export/home/phargrov/OMPI/openmpi-1.10.0rc3-solaris11-x86-ss12u4/BLD/opal/.libs -R/export/home/phargrov/OMPI/openmpi-1.10.0rc3-solaris11-x86-ss12u4/INST/lib -L/export/home/phargrov/OMPI/openmpi-1.10.0rc3-solaris11-x86-ss12u4/BLD/orte/.libs -L/export/home/phargrov/OMPI/openmpi-1.10.0rc3-solaris11-x86-ss12u4/BLD/opal/.libs ../../../../ompi/.libs/libmpi.so /export/home/phargrov/OMPI/openmpi-1.10.0rc3-solaris11-x86-ss12u4/BLD/orte/.libs/libopen-rte.so /export/home/phargrov/OMPI/openmpi-1.10.0rc3-solaris11-x86-ss12u4/BLD/opal/.libs/libopen-pal.so -lsendfile -lpicl -lkstat -llgrp -lpciaccess -lm -lsocket -lnsl -lc  -m32
ld: fatal: symbol 'ompi_sad_panda_' is multiply-defined:
        (file profile/.libs/libmpi_mpifh_pmpi.a(psizeof_f.o) type=FUNC; file ./.libs/libmpi_mpifh_sizeof.a(sizeof_f.o) type=FUNC);
```

However, if I revert the change to the perl script, but keep the edit (above) to the Makefile, then things build fine.
So, I believe the true problem is not the warned-about "empty" fortran file, but rather that the automake logic is not passing any objects to libtool.

There are also notable differences between mpif-h/Makefile.am and mpif-h/profile/Makefile.am, but I cannot tell if they have any significance here.  The later doesn't seem to be sandboxing the sizeof code into its own convenience lib as the former does.  I am wondering if there was perhaps an incomplete PR at some point?

jsquyres:
@PHHargrove Sigh.  Ok.  Let's try again -- instead of me pushing more commits to master that don't solve the problem, I've got a local branch with the next fix in it.  I applied it to v1.10 branch, and then made a tarball.  Translation: here's a v1.10 tarball that contains what I think might fix the problem -- can you give it a whirl?

http://www.open-mpi.org/~jsquyres/for-paul/

PHHargrove:
@jsquyres Testing the tarball now

PHHargrove:
@jsquyres 

Still not working:

```
  GENERATE psizeof_f.f90
  FC       psizeof_f.lo
  FCLD     libmpi_mpifh_psizeof.la
  CCLD     libmpi_mpifh_pmpi.la
  FCLD     libmpi_mpifh_sizeof.la
  CCLD     libmpi_mpifh.la
ld: fatal: file ./.libs/libmpi_mpifh_sizeof.a: open failed: No such file or directory
```

This looks like there is still an issue on mpif-h/Makefile.am, because here are some key lines from the corresponding generated Makefile:

```
-bash-$ grep nodist_libmpi_mpifh_sizeof_la Makefile 
#nodist_libmpi_mpifh_sizeof_la_OBJECTS = sizeof_f.lo
        $(nodist_libmpi_mpifh_sizeof_la_OBJECTS)
        $(nodist_libmpi_mpifh_sizeof_la_SOURCES)
#nodist_libmpi_mpifh_sizeof_la_SOURCES = sizeof_f.f90
```

Notice that the OBJECTS and SOURCES variables are commented out (and thus empty).
That explains the lack of "GENERATE" and "FC" lines for sizeof_f in the output at the start of this comment, and the lack of objects on the libtool command line I pointed out in email.

If I manually uncomment both lines, then things build just fine.
So I think there is still an issue in mpif-h/Makefile.am, though mpif-h/profile/Makefile.am is probably now correct.

**HOWEVER**, I am not sure this work you are doing is moving in the right direction by adding a dummy function to to an otherwise-empty source file to generate an almost-empty object to be linked into a almost-empty convenience library.
In the case that FC cannot implement sizeof, would it not make more sense to just **omit** libmpi_mpifh_sizeof.la from libmpi_mpifh_la_LIBADD?   I suspect that the following (or something like it) might be the right approach here:

```
--- ompi/mpi/fortran/mpif-h/Makefile.am.orig Fri Aug 14 10:44:47 2015
+++ ompi/mpi/fortran/mpif-h/Makefile.am     Fri Aug 21 10:45:13 2015
@@ -114,8 +114,8 @@
 if BUILD_FORTRAN_SIZEOF
 # Do not dist this file; it is generated
 nodist_libmpi_mpifh_sizeof_la_SOURCES = sizeof_f.f90
-endif
 libmpi_mpifh_la_LIBADD += libmpi_mpifh_sizeof.la
+endif

 sizeof_pl = $(top_srcdir)/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
```

Since I don't know for sure the conditions uder which BUILD_FORTRAN_SIZEOF is true, I cannot be certain this is the right fix.  Additionally, this doesn't prevent the attempts to build libmpi_mpifh_sizeof.la - it just prevents its _use_.  So, even _if_ the change above is correct and sufficient, it is not "complete".

jsquyres:
Arrgh... I see how I missed this.  I was developing on master and then porting my patches to v1.10.  The problem is that master already has exactly the fix you're talking about (have the LIBADD+= in the "if" conditional already), so I didn't see that v1.10 had the LIBADD+= _outside_ the conditional.

Ok, I'll adapt the PR for v1.10 to account for this as well.

FWIW, my final solution was to be double-defensive: not only does it ensure that there are now differently-named subroutines in mpif-h/ and mpif-h/profile, but those libraries shouldn't even be compiled at all (because of those "if" AM_CONDITIONALs).  I'll have a new tarball to test shortly.

PHHargrove:
@jsquyres 
OK, my pandas are lined up at the starting line awaiting a tarball...

jsquyres:
New v1.10 tarball (reflecting the new state of this PR) available at http://www.open-mpi.org/~jsquyres/for-paul/

PHHargrove:
@jsquyres I have begun testing...

PHHargrove:
:+1:

Was delayed slightly by running out of disk space.
I got good builds from Jeff's tarball for all four combinations {m32,m64} X {Studio,Gnu}

rhc54:
@PHHargrove THANK YOU, PAUL!

rhc54:
@jsquyres please go ahead and pull this when you have the PR branch ready (may already be so, but wanted to be sure)

PR #369, opened 2015-07-06
btl/openib: queue pending fragments once only when running out of credit

ggouaillardet:
Fixes open-mpi/ompi#640

(cherry picked from commit open-mpi/ompi@9f171de4125643345c605637356234499182ee6d)

PR #536, opened 2015-08-24
Java garbage collection bugfix

nrgraham23:
This pull request adds an arraylist of type Buffer to
the Request class.  Whenever a request object is created
that has associated buffers, the buffers should be added
to this array list so the java garbage collector does
not dispose of the buffers prematurely.

This is a more robust expansion on the idea first proposed by
@ggouaillardet

Fixes #369

Signed-off-by: Nathaniel Graham ngraham@lanl.gov

@hppritcha @jsquyres

:bot:milestone:v2.0.0
;bot:label:enhancement


hppritcha:
@jsquyres this is ready to go in

PR #538, opened 2015-08-24
Remove mx from the 1.10 release series

rhc54:
Revert "Correct the WRAPPER_EXTRA_LDFLAGS"

This reverts commit open-mpi/ompi-release@f5fd0999b0796d293514a8a16a4ad4c1cbab63b9.


jsquyres:
There's some mentions of MX in the README that also need to be removed.  Probably should also add a bullet in NEWS about how MX support has been removed, too.

PR #580, opened 2015-09-11
Fix cuda verbosity messages

rolfv:
(cherry picked from commit open-mpi/ompi@90dd1d264b9bab6c008fc95bb7460a167353445a)

Just a few cases where some output had gone stale.  Since @jsquyres has nothing better to do, I will give this to him.

### DIFF
--- a/opal/mca/common/cuda/common_cuda.c
+++ b/opal/mca/common/cuda/common_cuda.c
@@ -1103,9 +1103,9 @@ int cuda_getmemhandle(void *base, size_t size, mca_mpool_base_registration_t *ne
  */
 int cuda_ungetmemhandle(void *reg_data, mca_mpool_base_registration_t *reg)
 {
-    CUDA_DUMP_EVTHANDLE((100, ((mca_mpool_common_cuda_reg_t *)reg)->data.evtHandle, "cuda_ungetmemhandle"));
     opal_output_verbose(10, mca_common_cuda_output,
                         "CUDA: cuda_ungetmemhandle (no-op): base=%p", reg->base);
+    CUDA_DUMP_MEMHANDLE((100, ((mca_mpool_common_cuda_reg_t *)reg)->data.memHandle, "cuda_ungetmemhandle"));
 
     return OPAL_SUCCESS;
 }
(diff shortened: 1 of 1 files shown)


opal/mca/common/cuda/common_cuda.c
```diff
@@ -1691,9 +1691,9 @@ static void cuda_dump_memhandle(int verbose, void *memHandle, char *str) {
     }
     memcpy(&memH, memHandle, sizeof(memH));
     opal_output_verbose(verbose, mca_common_cuda_output,
-                        "%s:ctxId=%d, pid=%d, size=%d, blocksize=%d, offset=%d, gpuId=%d, "
+                        "%s:ctxId=%p, pid=%d, size=%d, blocksize=%d, offset=%d, gpuId=%d, "
```
jsquyres: Do you really want `%p` here?  `ctxId` is a `uint64_t`, so don't you want `PRIu64`, or `PRIx64`?

opal/mca/common/cuda/common_cuda.c
```diff
@@ -1718,8 +1715,8 @@ static void cuda_dump_evthandle(int verbose, void *evtHandle, char *str) {
     }
     memcpy(&evtH, evtHandle, sizeof(evtH));
     opal_output_verbose(verbose, mca_common_cuda_output,
-                        "CUDA: %s:ctxId=%d, pid=%d, index=%d",
-                        str, (int)evtH.ctxId, evtH.pid, (int)evtH.index);
+                        "CUDA: %s:pid=%d, serial=%d, index=%d",
+                        str, (int)evtH.pid, (int)evtH.serial, (int)evtH.index);
```
jsquyres: These vars are `unsigned long` -- can't they be truncated if you print them as `int`s?  Why is a PID an `unsigned long`, BTW?

If they are really `unsigned long`, then use the appropriate `%` sequence to print `unsigned long` -- don't squash them into an `int`.

rolfv:
All good points.  I have fixed them in a PR for the master.  If that looks good, then I will also squash them into this release one.  So, can you look at https://github.com/open-mpi/ompi/pull/893?

jsquyres:
(just putting in the xref here...)

Looks like open-mpi/ompi#893 is now good to go.

rolfv:
OK, I think I have this one right also now.
bot:assign: @jsquyres

hppritcha:
odd mlnx timeout
bot:retest

jsquyres:
:+1:

PR #581, opened 2015-09-15
btl/sm: mca_btl_sm_sendi: do not set *descriptor when descriptor is NULL

ggouaillardet:
(cherry picked from commit open-mpi/ompi@d5af5d106c12b836b4ae3984ac62ee5634ccdb65)


ggouaillardet:
:bot:assign: @hjelmn 
:bot:label:bug
:bot:milestone:v2.0.0

ggouaillardet:
@hjelmn i ran into this issue with master and random/ticket-1944-test4 from the ibm test suite.
so far, v2.x ran just fine without the fix ... since it is a harmless one and you requested it, here is the PR

hjelmn:
Weird that the problem does not occur on 2.x but it is a problem here since descriptor can be NULL.

:+1:

PR #590, opened 2015-09-20
Close only used file descriptors at fork/exec

rhc54:


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com:8888/jenkins/job/gh-ompi-release-v1.8-pr/723/ for details.

orte/mca/odls/default/odls_default_module.c
```diff
@@ -400,6 +403,41 @@ static void send_error_show_help(int fd, int exit_status,
...
+    }
+    DIR *dir = opendir(fds_dir);
+    free(fds_dir);
+    if (NULL == dir) {
+        return ORTE_ERR_FILE_OPEN_FAILURE;
+    }
+    struct dirent *files;
+    while (NULL != (files = readdir(dir))) {
+        if (0 == strncmp(files->d_name,".",1) ||
+            0 == strncmp(files->d_name,"..",2)) {
```
jsquyres: These two string comparisons are technically not correct.

A file named ".foo" will match them as well.  And ".." will match the first test, making the second test kinda meaningless.

I think the intent of this loop is to find only the files in the directory that are numbers.  Indeed, in /proc/pid/fd, that should be the only types of files that are in that dir.  But if we only want the numeric files, perhaps using isdigit() on the first character would be a better test...?

jsquyres: Should have mentioned @plesn on the original comment, sorry...

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com:8888/jenkins/job/gh-ompi-release-v1.8-pr/728/ for details.

jsquyres:
:+1:

orte/mca/odls/default/odls_default_module.c
plesn: I agree that isdigit() is better indeed !
And also I should have used "/proc/self/fd" instead of "/proc/$pid/fd" to avoid a useless heap allocation !

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com:8888/jenkins/job/gh-ompi-release-v1.8-pr/732/ for details.

PR #621, opened 2015-09-30
v1.10 - add mellanox platform specific patches

mike-dubman:


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/773/ for details.

jsquyres:
There was strong pushback against including organization/vendor-specific patches in community distributions on the call yesterday.

jladd-mlnx:
bot:label:critical

hppritcha:
I would like to explore some other options for these patches before giving up and including them as patches within the upstream ompi.

I'd like to have PRs opened on master to pull in at least patches 0001 and 0003.  I realize patch
0002 may be somewhat controversial, but I would also like to have a PR opened to master for some variant of this patch as well.   The patch is trying to address an issue which may potentially impact other vendors, particularly in those cases where the pmix barrier is relying on an external PMI implementation which has no potential for supporting hooks for an ompi progress callback function.   I think there are enough tools in ompi that we could come up with something more elegant than an explicit mpi barrier on MPI_COMM_WORLD near the beginning of the finalize procedure.  

The intent would be to push the PRs back to the 1.10.X branch as well as v2.x once accepted in to master.

As far as vendor patches,  I did not get pushback at LANL about the idea of having a patch infrastructure in place in Open MPI for which the sysadmin could then go to a vendor website and get the recommended patches to use as part of the build of the upstream Open MPI.  Would that approach be okay?  We'd want to add some robustness to the current patch support framework to insure that for customers, the modified build process is straightforward and reproducible, and that for vendors, they can determine exactly what patches were applied by the customer to the upstream Open MPI.

jsquyres:
@hppritcha I like your thinking here:
1. Let's work these three patches and get them into master, v1.10, and v2.x directly (ASAP).  Then the immediate problem goes away, and we can release v1.10.1 with what Mellanox needs, and what others will find acceptable.
2. Then we can spend a little time and think about how we want to support scenarios like you / Mellanox describe for future cases like this.  We can probably come up with a mutually-agreeable solution, but it may take a little thought + work.

mike-dubman:
sounds good! thanks for help!

mike-dubman:
no need, replaced by #631 and #632

PR #493, opened 2015-08-12
Fix PMI and PMI2 builds

nysal:
(cherry picked from commit open-mpi/ompi@2d2ea63)

PR #636, opened 2015-10-05
v2.x - Add in-finalize indicator, fca fall back to prev barrier if in-finalize

mike-dubman:
master commit: https://github.com/open-mpi/ompi/pull/972

### DIFF
--- a/ompi/runtime/mpiruntime.h
+++ b/ompi/runtime/mpiruntime.h
@@ -55,6 +55,8 @@ OMPI_DECLSPEC extern bool ompi_mpi_initialized;
 OMPI_DECLSPEC extern bool ompi_mpi_finalized;
 /** Has the RTE been initialized? */
 OMPI_DECLSPEC extern bool ompi_rte_initialized;
+/** Did mpi start to finalize? */
+OMPI_DECLSPEC extern int32_t ompi_mpi_finalize_started;
 
 /** Do we have multiple threads? */
 OMPI_DECLSPEC extern bool ompi_mpi_thread_multiple;
--- a/ompi/mca/coll/fca/coll_fca_ops.c
+++ b/ompi/mca/coll/fca/coll_fca_ops.c
@@ -153,6 +153,10 @@ int mca_coll_fca_barrier(struct ompi_communicator_t *comm,
     int ret;
 
     FCA_VERBOSE(5,"Using FCA Barrier");
+    if (OPAL_UNLIKELY(ompi_mpi_finalize_started)) {
+        FCA_VERBOSE(5, "In finalize, reverting to previous barrier");
+        goto orig_barrier;
+    }
     ret = fca_do_barrier(fca_module->fca_comm);
     if (ret < 0) {
         if (ret == -EUSEMPI) {
--- a/ompi/runtime/ompi_mpi_finalize.c
+++ b/ompi/runtime/ompi_mpi_finalize.c
@@ -93,7 +93,6 @@ extern bool ompi_enable_timing_ext;
 int ompi_mpi_finalize(void)
 {
     int ret;
-    static int32_t finalize_has_already_started = 0;
     opal_list_item_t *item;
     ompi_proc_t** procs;
     size_t nprocs;
@@ -106,7 +105,7 @@ int ompi_mpi_finalize(void)
        ompi_comm_free() (or run into other nasty lions, tigers, or
        bears) */
 
-    if (! opal_atomic_cmpset_32(&finalize_has_already_started, 0, 1)) {
+    if (! opal_atomic_cmpset_32(&ompi_mpi_finalize_started, 0, 1)) {
         /* Note that if we're already finalized, we cannot raise an
            MPI exception.  The best that we can do is write something
            to stderr. */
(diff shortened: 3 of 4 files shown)


mike-dubman:
:bot:assign: @hppritcha 
:bot:label:bug
:bot:milestone:v2.0.0

ompi/runtime/mpiruntime.h
```diff
@@ -55,6 +55,8 @@ OMPI_DECLSPEC extern bool ompi_mpi_initialized;
 OMPI_DECLSPEC extern bool ompi_mpi_finalized;
 /** Has the RTE been initialized? */
 OMPI_DECLSPEC extern bool ompi_rte_initialized;
+/** Did mpi start to finalize? */
+OMPI_DECLSPEC extern int32_t ompi_mpi_finalize_started;
```
jsquyres, line 59: Just curious -- why int32 instead of bool?

mike-dubman, line 59: - similar functionality existed before this patch and "mpi_has_already_started" variable was local and int32_t. (see original before patch)
- the patch just externalized it, w/o changing logic or data layout.

``` shell
diff --git a/ompi/runtime/ompi_mpi_finalize.c b/ompi/runtime/ompi_mpi_finalize.c
index 51274992..7935ebeb 100644
--- a/ompi/runtime/ompi_mpi_finalize.c
+++ b/ompi/runtime/ompi_mpi_finalize.c
@@ -93,7 +93,6 @@ extern bool ompi_enable_timing_ext;
 int ompi_mpi_finalize(void)
 {
     int ret;
-    static int32_t finalize_has_already_started = 0;
     opal_list_item_t *item;
     ompi_proc_t** procs;
     size_t nprocs;
@@ -106,7 +105,7 @@ int ompi_mpi_finalize(void)
        ompi_comm_free() (or run into other nasty lions, tigers, or
        bears) */

-    if (! opal_atomic_cmpset_32(&finalize_has_already_started, 0, 1)) {
```
- It probably int32 because this variable was used for refcnt and allowed multiple invocation of finalize.

jsquyres, line 59: The other initialized / finalized types are bool's -- might as well make this one the same...?

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/800/ for details.

ompi/runtime/mpiruntime.h
mike-dubman, line 59: yep, we did a lot of testing w/ this patch and tried to keep it as close as possible to the original code.
Maybe we can  add type change in master only (for safety) and keep release branches w/ minimal refactoring changes?

jsquyres, line 59: It's probably 6 of one, half dozen of the other.  I minorly prefer keeping all the types together on all branches.  It does seem a little weird to commit something that we know is not what we want in the final state.  Is there much additional QA burden to change from int32_t to bool?

hjelmn:
@jsquyres Note the code use atomics to set finalize_has_already_started. We do not have atomics for boolean types.

jsquyres:
That's a good reason.  Should this PR then be expanded to do the same thing for `ompi_mpi_init()`?

mike-dubman:
do you mean to externalize various flags, like:

``` shell
bool ompi_mpi_init_started = false;
bool ompi_mpi_initialized = false;
bool ompi_mpi_finalized = false;
bool ompi_rte_initialized = false;
int32_t ompi_mpi_finalize_started = false;
```
- they are already externalized.
- Or you mean to convert them from bool to int32? Don`t think it is needed, because they usage is correct (only assignment of true/false).

jsquyres:
No, I mean: 
1. should ompi_mpi_init_started be converted to int32_t and assigned via atomic
2. check for ompi_mpi_init_started==1 (via atomic) and fail out the way ompi_mpi_finalize does

@bosilca is also advocating that INITIALIZED and FINALIZED be adapted to check for the `_started` values, as well (which would work well if they're set/tested via atomics) in #493.  That work should obviously be different than this work, but this work would nicely compliment/enable it (i.e., the set-by-atomic part).

hppritcha:
Can we switch to using atomics rather than boolean?  Per discussions at devel call today these could then be more easily reused for open-mpi/ompi#493.

jsquyres:
@hppritcha That's what I advocated in the prior comment :smile:

hppritcha:
I was just adding my vote in for the change.

hppritcha:
@miked-mellanox could you switch to these variables to using atomic updates?  This would help with addressing issue open-mpi/ompi#493

mike-dubman:
@hppritcha - yep, could you please open a ticket on me and will handle it.

we will do it in separate PR for master and v2.x as this PR and requested change are logically unrelated.

jsquyres:
Actually, I worked up the idea of updating INIT/FINALIZE/etc. over the weekend in open-mpi/ompi#1007, and it turns out that converting all these variables to use atomics wouldn't solve the problem, anyway, so let's let this PR go through as-is, and if we want to convert some/all of them to atomics (or all non-atomics), then we'll handle that separately.

jsquyres:
@hppritcha So I think this PR is now good to go: :+1:

PR #657, opened 2015-10-08
opal/mpool: add support for passing access flags to register

hjelmn:
This commit adds a access_flags argument to the mpool registration
function. This flag indicates what kind of access is being requested:
local write, remote read, remote write, and remote atomic. The values
of the registration access flags in the btl are tied to the new flags
in the mpool. All mpools have been updated to include the new argument
but only the grdma and udreg mpools have been updated to make use of
the access flags. In both mpools existing registrations are checked
for sufficient access before being returned. If a registration does
not contain sufficient access it is marked as invalid and a new
registration is generated.

:bot:assign: @bosilca 
:bot:label:enhancement
:bot:milestone:v2.0.0

(cherry-picked from open-mpi/ompi@59aa93e1b6b7e9a3548c4620e9cf031cfc608814)

### DIFF
--- a/opal/mca/btl/btl.h
+++ b/opal/mca/btl/btl.h
@@ -247,28 +247,29 @@ typedef uint8_t mca_btl_base_tag_t;
 #define MCA_BTL_ERROR_FLAGS_NONFATAL 0x2
 #define MCA_BTL_ERROR_FLAGS_ADD_CUDA_IPC 0x4
 
-/** registration flags */
+/** registration flags. the access flags are a 1-1 mapping with the mpool
+ * access flags. */
 enum {
     /** Allow local write on the registered region. If a region is registered
      * with this flag the registration can be used as the local handle for a
      * btl_get operation. */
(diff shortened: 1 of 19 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/830/ for details.

bosilca:
My only comment about this patch is the fact that now we have 2 sets of (extremely similar) flags, one in the BTL and one in the MPOOL. Can we replace the BTL flags with the MPOOL ones?

hjelmn:
@bosilca The BTL registration flags that overlap the mpool access flags are mapped 1-1 to the mpool access flags. The last flag (MCA_BTL_REG_FLAG_CUDA_GPU_MEM) is similar to an mpool flag.

I am planning on rewriting the mpool and rcache in the next month. I will look into de-duplicating flags as part of that work.

bosilca:
In this case no more comments. :+1:

jsquyres:
@hppritcha Good to go.

PR #668, opened 2015-10-14
Patch configure to support PowerPC little endian

nysal:
Open MPI v1.10 tarballs are generated with libtool 2.4.2
The earliest libtool release that supports PowerPC little endian
is 2.4.3. Patch configure to enable support for this architecture.
See open-mpi/ompi/issues/396 for details.

:bot:assign: @jsquyres 
:bot:milestone:v1.10.1
:bot:label:enhancement

### DIFF
--- a/autogen.pl
+++ b/autogen.pl
@@ -4,6 +4,8 @@
 # Copyright (c) 2010      Oracle and/or its affiliates.  All rights reserved.
 # Copyright (c) 2013      Mellanox Technologies, Inc.
 #                         All rights reserved.
+# Copyright (c) 2015      IBM Corporation.  All rights reserved.
+#
 # $COPYRIGHT$
 # 
 # Additional copyrights may follow
@@ -179,7 +181,6 @@ sub process_subdir {
         print "--- Found configure.in|ac; running autoreconf...\n";
         safe_system("autoreconf -ivf");
         print "--- Patching autotools output... :-(\n";
-        patch_autotools_output($start);
     } else {
         my_die "Found subdir, but no autogen.sh or configure.in|ac to do anything";
     }
@@ -187,6 +188,9 @@ sub process_subdir {
     # Ensure that we got a good configure executable.
     my_die "Did not generate a \"configure\" executable in $dir.\n"
         if (! -x "configure");
+
+    # Fix known issues in Autotools output
+    patch_autotools_output($start);
 
     # Chdir back to where we came from
     chdir($start);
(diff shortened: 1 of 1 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/843/ for details.

jsquyres:
This should go on master, too, right?

We probably don't _need_ it on master (i.e., it's only for the case where someone using an older Libtool -- and not on a Power machine with the adjustment shell script -- makes a tarball for a Power user), but  if autogen.pl is otherwise the same between v1.10 and master, it might be good to keep them in sync.

...I just checked, a diff between master autogen.pl and v1.10 autogen.pl shows that they are identical.  We should probably keep autogen.pl identical between master, v1.10, and v2.x.

:+1:

nysal:
A minor update to the patch - Added a copyright statement, tweaked the verbose message and moved invocation of patch_autotools_output to the same location as master branch. I'll prepare a similar patch for master and v2.x branches. Thanks for the review @jsquyres

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/851/ for details.

Zhiming-Wang:
@nysal
If the later version libtool which the bug is fixed will be used for master, there will be two cases for Power LE like this. Right, but not elegant. 

Such as "configure" 

> ... ...
>            ppc*le-*linux*|powerpc*le-*linux*)
>              LD="${LD-ld} -m elf64lppc"
>              ;;
>            powerpcle-*linux*)
>               LD="${LD-ld} -m elf64lppc"
>              ;;
>            powerpc-*linux*)
>              LD="${LD-ld} -m elf64ppc"
>              ;;
> ... ...

jsquyres:
@Zhiming-Wang Good catch!  It looks like @nysal updated to only patch if necessary.

nysal:
RHEL is the odd one out, the patch fails as expected on the other LE platforms. The reason the patch misbehaves on RHEL is that their fix for LE in libtool.m4 is a slight departure from how its fixed in upstream libtool. Hence our regex unfortunately matches and incorrectly tries to apply the patch. So the fix I would like to try is to test for the string 'powerpc64le-_linux_)' and if there is a match do not try to apply the patch. This will work on RHEL and SLES. However Ubuntu has "powerpc64le-*)", so the test fails and we will try to apply the patch. But the regex will not match and hence patch is not applied. So we are good (theoretically) in all 3 cases.
@Zhiming-Wang is testing the modified fix

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/854/ for details.

nysal:
@Zhiming-Wang has confirmed that it works on RHEL, Ubuntu and SLES. So this seems good to go for 1.10.1.

Zhiming-Wang:
@nysal  @jsquyres 
Yes, it worked.

jsquyres:
@rhc54 This is good to for go for v1.10.1.

PR #675, opened 2015-10-15
btl/ugni: fix very poor aries bw problem

hppritcha:
The handling of RDMA get alignment in ugni BTL for Aries
(cray xc) was wrong, resulting in very poor bandwidth
for ugni BTL on aries.

Verified using osu_bw now gives sensible bandwidth on
Aries.

Fixes open-mpi/ompi#1005

@hjelmn  please review

(cherry picked from commit open-mpi/ompi@eaba98ce5d6f31f18437d1e4e36b7d73e422e854)

### DIFF
--- a/opal/mca/btl/ugni/btl_ugni_component.c
+++ b/opal/mca/btl/ugni/btl_ugni_component.c
@@ -231,6 +231,9 @@ btl_ugni_component_register(void)
     /* determine if there are get alignment restrictions */
     GNI_GetDeviceType (&device_type);
 
+    /*
+     * see def. of ALIGNMENT_MASK to figure this one out
+     */
     if (GNI_DEVICE_GEMINI == device_type) {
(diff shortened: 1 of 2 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/858/ for details.

hjelmn:
:+1:

hjelmn:
Before this patch we were falling back on the RDMA put protocol. Not sure why it was performing so poorly. I might have to take a look at that at some point.

hppritcha:
Actually I don't think it was using put path.

hjelmn:
Which is a little odd. If get fails (as it was here) it should fall back on RDMA put and only on send/recv if put fails. The numbers certainly look like send/recv numbers though.... Definitely a bug somewhere. Not critical when RDMA get is working but worth tracking down. I will try to track it down in the next week or so.

PR #686, opened 2015-10-19
coll/base: fix memory allocation in mca_coll_base_alltoall_intra_basi…

ggouaillardet:
…c_inplace

(cherry picked from commit open-mpi/ompi@0f23037775d215e021d4631ac2d43bcb02ee9a3e)

### DIFF
--- a/ompi/mca/coll/base/coll_base_alltoall.c
+++ b/ompi/mca/coll/base/coll_base_alltoall.c
@@ -46,7 +46,7 @@ mca_coll_base_alltoall_intra_basic_inplace(const void *rbuf, int rcount,
     MPI_Request *preq;
     char *tmp_buffer;
     size_t max_size;
-    ptrdiff_t ext;
+    ptrdiff_t ext, true_lb, true_ext;
 
     /* Initialize. */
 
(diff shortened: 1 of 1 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/870/ for details.

ggouaillardet:
:bot:assign: @bosilca 
:bot:label:bug
:bot:milestone:v2.0.0

jsquyres:
@bosilca gave this a :+1: in #687.

bosilca:
Actually, I was holding on this one, because I think the whole allocation of memory for collective operations is wrong. But, I need a little bit more time to think about

jsquyres:
@bosilca Doh.  Should I revert?

bosilca:
Leave it as is by now. Once I have a better understanding of the issue, I
will try to propose a cleaner approach.

bosilca:
The more I look at this patch, the more confused I get. The only thing this change does is slightly reducing the allocated memory. However, now the allocated memory and the max_size are out of sync, as one includes the gap while the other doesn't. In addition the allocated buffer is not consistent with the amount of data allocated, as it has not been moved back by the true_lb (as in the other functions).

bosilca:
I just pushed a PR to clean the handling of temporary buffers (mainly in collectives). Please take a look @ open-mpi/ompi#1047. However, it is just a bandaid, the root of all these problems still exists.

ggouaillardet:
my patch does **change** the size of the allocated memory.
if extent < true_extent (in Fortran, think of a matrix row), the allocated memory is slightly **increased**
i agree this patch does not handle non-zero `true_lb`.
i will review open-mpi/ompi#1047 from now

bosilca:
You're right, in the case where the extent was altered by a resize you used to formula that gives the right span of the data.

PR #688, opened 2015-10-19
Fix singleton operations when running under a SLURM allocation.

rhc54:
Sadly, SLURM's PMI will return success even if the PMI server isn't actually available. This leads to erroneous selection of pmix and ess components. So add a further requirement (namely, that we see a job_step envar) to the SLURM pmix components along with some modification of ess selection code to avoid the problem

(cherry picked from commit open-mpi/ompi@363f62a506b3da531dc24526d36d46e3e97fb2d2)

### DIFF
--- a/opal/mca/pmix/pmix1xx/pmix/src/client/pmix_client.c
+++ b/opal/mca/pmix/pmix1xx/pmix/src/client/pmix_client.c
@@ -239,6 +239,11 @@ int PMIx_Init(pmix_proc_t *proc)
         return PMIX_SUCCESS;
     }
 
+    /* if we don't see the required info, then we cannot init */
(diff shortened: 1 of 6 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/872/ for details.

hppritcha:
I want to double check this in an alps env.

hppritcha:
:+1: works for me!

hppritcha:
@jsquyres I think this is ready to go if @rhc54 is good with it

hppritcha:
@jsquyres this is good to go.

PR #713, opened 2015-10-23
Silence a couple of warnings from valgrind and compilers

rhc54:
Since some pmix components may return success with a NULL value from a "get", check for that situation before attempting to unload the data. Preset the hostname before calling modex_recv to get it so unload properly checks for NULL. Cast a returned value to the correct ompi_proc_t pointer

(cherry picked from commit open-mpi/ompi@4c12022a50efb68ea116a3ffeadeb2384bda9024)


jsquyres:
:+1:

PR #716, opened 2015-10-24
Patch configure to support PowerPC little endian

nysal:
The earliest libtool release that supports PowerPC little endian
is 2.4.3. Patch configure to enable support for this architecture.

(cherry picked from commit open-mpi/ompi@12c5063d58d3)

### DIFF
--- a/autogen.pl
+++ b/autogen.pl
@@ -7,6 +7,8 @@
 # Copyright (c) 2013-2014 Intel, Inc.  All rights reserved.
 # Copyright (c) 2015      Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
+# Copyright (c) 2015      IBM Corporation.  All rights reserved.
+#
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


nysal:
:bot:assign: @jsquyres 
:bot:label:enhancement
:bot:milestone:v2.0.0

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/919/ for details.

gpaulsen:
v2.0 is using libtool 2.4.6, and should not need this patch.  We talked about both removing the autogen.pl from the tar ball (a general enhancement), and a specific idea to check that the version of libtool is new enough on Power Machines, and die more gracefully.

jsquyres:
FYI: the table of which autotools are used for each tarball series are listed here http://www.open-mpi.org/source/building.php

nysal:
Please see https://github.com/open-mpi/ompi-release/pull/668 for the background. In summary, Jeff suggested we should probably keep autogen.pl as identical as possible across v1.10, v2.x and master. Currently autogen.pl checks if libtool version is >= 2.4.2. The autotools version used for generating the official tarball is 2.4.6, which is fine. But unless we change the minimum supported libtool version in autogen.pl, we might still need this patch.

jsquyres:
@gpaulsen @nysal Yeah, I did suggest that (keeping them in sync).  I think I said the opposite on this past Tuesday webex -- but upon thinking about it further, I think there's value in keeping them in sync.

:+1:

jsquyres:
@hppritcha Good to go

PR #727, opened 2015-10-28
Cleanup the PMIx direct modex support

rhc54:
Add an MCA parameter pmix_base_async_modex that will cause the async modex to be used when set to 1. Default it to 0 for now to continue current default behavior.

Also add an MCA param pmix_base_collect_data to direct that the blocking fence shall return all data to each process. Obviously, this param has no effect if async_modex is used.

(cherry picked from commit open-mpi/ompi@267ca8fcd3a59b780491d80d29e870061d8dac56)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-v1.8-pr/939/ for details.

jsquyres:
I can't really verify the giant new chunk of code in pmix_server_get.c and the big deletion from pmix_server_ops.c, but the rest of it looks ok...?

hppritcha:
since this is an opt in it sounds low risk.  Still I will see what happens if I try to use async modex with a BTL which is purported to be able to make use of it.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/964/ for details.

PR #747, opened 2015-11-07
Need to delay registration of the waitpid callback until after the fo…

rhc54:
…rk/exec of the child process. Fix the bit testing of process type so that the proper state component gets selected for HNP.

(cherry picked from commit open-mpi/ompi@f1483eb2dc2a3f53f072eb6383a2d1c1537257c4)

PR #753, opened 2015-11-10
pmix: fix finalize ordering issue

jsquyres:
Don't destroy the event base before destroying all objects that contain events (and call event_del(), which uses the event base on which event is enqueued).

This matches the ordering in the Open MPI master.

@rhc54 How do you want to handle this?  Do you want piecemeal PMIx bug fixes like this, or do you want to do a re-sync of PMIx from OMPI master (or PMIx master?)?

This issue was initially found by valgrind on https://github.com/open-mpi/ompi-release/pull/747#issuecomment-155152968, and then had its own issue opened in open-mpi/ompi#1117.

@rhc54 Please review.

### DIFF
--- a/opal/mca/pmix/pmix1xx/pmix/src/client/pmix_client.c
+++ b/opal/mca/pmix/pmix1xx/pmix/src/client/pmix_client.c
@@ -5,6 +5,7 @@
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2014      Artem Y. Polyakov <artpol84@gmail.com>.
  *                         All rights reserved.
+ * Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


rhc54:
As I said on the other ticket you opened on this, it appears to be simply a case of 2.x not having all the pmix updates yet. I wouldn't piecemeal it - the plan is to provide a complete update once pmix 1.1.0 is released.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/999/ for details.

jsquyres:
Per discussion on the 2015-11-10, @rhc54 says the fix is fine.  The ultimate plan is to re-sync pmix in both master and v2.x when pmix 1.1 is released (at some point in the near future).  So let's get this fix in.

PR #763, opened 2015-11-12
rpm: fix openmpi.spec not to include the /usr directory

ggouaillardet:
/usr cannot be included on RHEL7 like distros

(back ported from commit open-mpi/ompi@6ab32895825ed7ecf8bb7563f75a27962fa6b426)

### DIFF
--- a/contrib/dist/linux/openmpi.spec
+++ b/contrib/dist/linux/openmpi.spec
@@ -12,6 +12,8 @@
 # Copyright (c) 2006-2014 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2013      Mellanox Technologies, Inc.
 #                         All rights reserved.
+# Copyright (c) 2015      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 # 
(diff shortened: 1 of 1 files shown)


ggouaillardet:
:bot:assign: @jsquyres 
:bot:label:bug
:bot:milestone:v1.10.2

@rhc54, we already got two reports on the ML (specific to RHEL7 like), so i set the milestone to v1.10.2

mellanox-github:
Test FAILed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1015/ for details.

jsquyres:
:+1: 

@rhc54 I agree with @ggouaillardet -- we did get some mailing list bug reports about this.  This is just a packaging fix, so the risk is fairly low for v1.10.2.  FWIW, we've released packaging fixes (specifically: new SRPMs) without doing a formal release before.  That's also an option.  But it's also not a huge tragedy if it's not included in 1.10.2, given the very short timeframe / high degree of risk -- not very many users use the Open MPI SRPM.  So this one could go either way.

rhc54:
I agree this is close to zero risk

PR #772, opened 2015-11-18
correctly handle mpirun --host <user>@<ip>

ggouaillardet:
This is a one-off patch for the v1.10 series

Thanks Federico Reghenzani for reporting this issue

### DIFF
--- a/orte/util/nidmap.c
+++ b/orte/util/nidmap.c
@@ -326,11 +326,19 @@ int orte_util_encode_nodemap(opal_byte_object_t *boptr, bool update)
         }
         /* pack the name of the node */
         if (!orte_keep_fqdn_hostnames) {
-            nodename = strdup(node->name);
+            /* handle nodename is <login>@<host>
+             * the real nodename is <host> */
+            char * real_nodename = nodename = strdup(node->name);
+            real_nodename = strchr(nodename, '@');
+            if (NULL != real_nodename) {
+                real_nodename++;
+            } else {
+                real_nodename = nodename;
+            }
             /* if the nodename is an IP address, do not mess with it! */
-            if (!opal_net_isaddr(nodename)) {
+            if (!opal_net_isaddr(real_nodename)) {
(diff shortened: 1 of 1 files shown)


ggouaillardet:
:bot:milestone:v1.10.2
:bot:label:bug
:bot:assign: @rhc54

ggouaillardet:
see http://www.open-mpi.org/community/lists/users/2015/11/28042.php

as a side note, i have no idea why it works with two hosts (!)

mellanox-github:
Test FAILed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1034/ for details.

lanl-ompi:
Test FAILed.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1035/ for details.

rhc54:
@ggouaillardet Thank you for looking into this. I'm trying to understand it a bit better, but my limited time may be causing me to miss how this fixes the ssh. I assume that the first ssh from mpirun to the first node is working okay as the rsh/plm component properly parsed the username/addr. I'm guessing that the nidmap the new daemon was given had an incorrect address for the next one to be launched due to this bug, and that is what broke the tree-based launch.

Does that match what you found?

ggouaillardet:
@rhc54 as I wrote earlier, I do not understand everything ...
let's say that opal_net_is_addr("xxx@1.2.3.4") invokes getaddrinfo("xxx@1.2.3.4") and that is an undetermined behavior, or that this code path is only executed when more than two hosts are requested.
if it returns false, then xxx@1.2.3.4 is trimmed to xxx@1 and that will cause ssh to fail.
my pr ensures we invoke opal_net_is_addr("1.2.3.4") which returns true, and everything is fine.
makes (a bit more) sense ?

rhc54:
Okay, I walked thru the code and figured out what was happening - thanks for the fix!

PR #792, opened 2015-12-01
orterun.c: fix parse of user environment variables

stefano-garzarella:
When I launch mpirun, I need to pass some (ompss/nanos++) environment variable
like NX_ARGS in the following way:
export NX_ARGS="--verbose --cluster --cluster-network=mpi"
mpirun -np 2 -hostfile myhostfile -x NX_ARGS -x LD_PRELOAD testapp

I saw that NX_ARGS is not passed and I found the problem in the orte/tools/orterun/orterun.c file.
In line 1739, after use getenv(), you check if there is '=' in the string. If it's true,
you assumed that the string is in the form 'param'='value', but in this case fails, because
getenv() POSIX specification always returns a pointer to the value
in the environment, without "param=".

getenv() POSIX specification always returns a pointer to the value
in the environment.

### DIFF
--- a/orte/tools/orterun/orterun.c
+++ b/orte/tools/orterun/orterun.c
@@ -1736,17 +1736,11 @@ static int create_app(int argc, char* argv[],
             } else {
                 value = getenv(param);
                 if (NULL != value) {
-                    if (NULL != strchr(value, '=')) {
-                        opal_argv_append_nosize(&app->env, value);
-                        /* save it for any comm_spawn'd apps */
-                        opal_argv_append_nosize(&orte_forwarded_envars, value);
-                    } else {
                         asprintf(&value2, "%s=%s", param, value);
                         opal_argv_append_nosize(&app->env, value2);
                         /* save it for any comm_spawn'd apps */
                         opal_argv_append_nosize(&orte_forwarded_envars, value2);
                         free(value2);
-                    }
                 } else {
                     opal_output(0, "Warning: could not find environment variable \"%s\"\n", param);
                 }


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1074/ for details.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1075/ for details.

PHHargrove:
My recollection (and I can dig up supporting docs is necessary) is that pre-POSIX implementations of getenv() could optionally include the "param=" prefix in the result.  So, for maximum portability the code removed by the PR should _instead_ check explicitly for "param=".  However the code which assumes that the first '=' (if any) follows 'param' is simply wrong.

-Paul

PHHargrove:
I was able to follow up with the various BSD and Solaris manpages.
What I found does _not_ match by "recollection" in the previous comment.
What I found is that some implementations allow for a trailing '=' in the _argument_ to getenv().
No version I could find docs for is returning a value prefixed with 'param='.
So, I believe the PR is correct as it stands now (w/o any changes suggested in my previous comment).

-Paul

rhc54:
@PHHargrove Thanks for the review, Paul.

@stefano-garzarella This needs to come into our master first, and then can be ported to the release branch after it has been tested there. I'll move it to the appropriate place.

rhc54:
Actually, the master has a completely different code block for performing this function, and it looks correct. So this appears to be something isolated to the 1.10 branch, and I'll accept it as-is.

@stefano-garzarella Thanks!

stefano-garzarella:
@PHHargrove I'm agree with you! I also checked several manpages and I noticed that all the implementations respect the POSIX specification.

@rhc54 Thanks for merge :)

PR #878, opened 2015-12-28
plm/alps: only use srun for Native SLURM

hppritcha:
Turns out that the way the SLURM plm works
is not compatible with the way MPI processes
on Cray XC obtain RDMA credentials to use
the high speed network.  Unlike with ALPS,
the mpirun process is on the first compute
node in the job.  With the current PLM launch
system, mpirun (HNP daemon) launches the MPI
ranks on that node rather than relying on
srun.

This will probably require a significant amount
of effort to rework to support Native SLURM
on Cray XC's.  As a short term alternative,
have the alps plm (which gets selected by default
again on Cray systems regardless of the launch system)
check whether or not srun or alps is being used on the
system.  If alps is not being used, print a helpful
message for the user and abort the job launch.

@hjelmn please review

(cherry picked from commit open-mpi/ompi@39367ca0bf56a316682d2e1a690fb50d3ab6f244)

### DIFF
--- a/orte/mca/plm/alps/help-plm-alps.txt
+++ b/orte/mca/plm/alps/help-plm-alps.txt
@@ -39,3 +39,7 @@ the map for this application. This can be caused by a lack of
 an allocation, or by an error in the Open MPI code. Please check
 to ensure you have a ALPS allocation. If you do, then please pass
 the error to the Open MPI user's mailing list for assistance.
+#
+[slurm-not-supported]
+mpirun is not a supported launcher on Cray XC using Native SLURM.
+srun must be used to launch jobs on these systems.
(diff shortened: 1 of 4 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1204/ for details.

hjelmn:
:+1:

hppritcha:
unfortunately it looks like uh server is unreachable, ignoring the distcheck and disable-dlopen check

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

PR #883, opened 2016-01-05
fixes for the plfs file system

edgargabriel:
@hppritcha do you have somewhere access to a plfs installation? If not, let me know and I can set up something for you to test. Otherwise not sure how to review this pr, since without plfs it will not even compile the parts that were modifiied.

### DIFF
--- a/ompi/mca/fbtl/plfs/fbtl_plfs_preadv.c
+++ b/ompi/mca/fbtl/plfs/fbtl_plfs_preadv.c
@@ -28,74 +28,19 @@
 ssize_t  mca_fbtl_plfs_preadv (mca_io_ompio_file_t *fh )
 {
 
-    Plfs_fd *pfd = NULL;
+    Plfs_fd *pfd = fh->f_fs_ptr;
     plfs_error_t plfs_ret;
-    pfd = fh->f_fs_ptr;
     ssize_t total_bytes_read=0;
-
-    int i, block=1;
-    struct iovec *iov = NULL;
-    int iov_count = 0;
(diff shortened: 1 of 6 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1208/ for details.

hppritcha:
I am not currently set up for plus testing. Maybe we should at least install headers etc on uh Jenkins node.  I'm travel this week.  I will investigate next week.

edgargabriel:
ok, lets discuss that next week. I could install the software necessary for plfs, pvfs2, lustre etc. and we could have at least a compile time test that runs regularly.

edgargabriel:
@hppritcha I added a PLFS installation directory on ompi.cs.uh.edu, it is in /opt/PLFS. In theory, OpenMPI could now be configured with --with-plfs=/opt/PLFS and the component should at least compile correctly.

hppritcha:
@edgargabriel I'm asking the a couple of the people in my group here at LANL - this is something from LANL originally - about where I can actually test plfs fs.  What caused you to want to fix plfs in ompiio?  Was there a user reported problem?

edgargabriel:
yes, a user complained that ompio would not compile if plfs was 
detected. This fixes the compilation problem in that scenario.

hppritcha:
works for me.  There don't appear to be any PLFS file systems available here at LANL - the home of PLFS, so I'm assuming that just a compile is good enough.

:+1:

hppritcha:
@jsquyres ready to go

PR #893, opened 2016-01-13
v2.0.0: opal hotel: only delete events that have not yet fired

jsquyres:
The eviction callback, for convenience (and to avoid code duplication), use to call opal_hotel_checkout().  However, opal_hotel_checkout() deletes the eviction event -- which is fine to do when opal_hotel_checkout() is invoked by the application.  But when it's invoked by the same event that it's deleting, it can cause Bad Things to happen.

For simplicity, instead of invoking opal_hotel_checkout() from the eviction callback, just duplicate the checkout logic into the eviction callback function (and skip the delete-the-evict-event part).

For good measure, put a comment in all three places where the checkout logic occurs (because it's inlined): don't change this logic without changing all 3 places.

Finally, also add a line in the docs for opal_hotel_init() warning users from calling opal_hotel_checkout() from their eviction callback.

(cherry picked from commit open-mpi/ompi@270cc1115679a0b44a998a0555096db05f35b453)

@rhc54 please review

### DIFF
--- a/opal/class/opal_hotel.c
+++ b/opal/class/opal_hotel.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012      Cisco Systems, Inc. All rights reserved.
+ * Copyright (c) 2012-2016 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2012      Los Alamos National Security, LLC. All rights reserved
  * Copyright (c) 2015      Intel, Inc. All rights reserved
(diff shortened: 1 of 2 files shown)


rhc54:
:+1:

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1218/ for details.

jsquyres:
@hppritcha good to go

PR #902, opened 2016-01-20
mpi: have MPI_Wtick() return the period (and not the frequency) if OP…

ggouaillardet:
…AL_TIMER_CYCLE_NATIVE

(cherry picked from commit open-mpi/ompi@2adbe273d66b20582eda34f85d54496da89178d7)

### DIFF
--- a/ompi/mpi/c/wtick.c
+++ b/ompi/mpi/c/wtick.c
@@ -10,7 +10,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2007-2014 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1235/ for details.

ggouaillardet:
:bot:label:bug
:bot:milestone:v2.0.0
:bot:assign: @bosilca

bosilca:
Good catch. :+1:

hppritcha:
@jsquyres this looks good to me.

jsquyres:
@ggouaillardet @bosilca How serious is this? Should it go to v1.10, too?

@rhc54 Heads up on a (small) potential 1.10.2 issue...

bosilca:
Based on the fact that we had no complains about this for so long I would guess very few users run with native timers. But I think it is a good idea to move this minimal patch to 1.10.

jsquyres:
Thinking about this a little more...

"Native timers" is any Linux x86-64 platform these days, right?  If so, that's the majority of our userbase.

I agree that most people probably don't use timers -- but perhaps we should reserve this change in behavior for v2.0.0?  (I know, I was the one who asked about it in the first place :smile: )

ggouaillardet:
in v1.10, the bug can only occur if **WINDOWS** is defined
so only Cygwin could be affected

the error impacted the c++ test suite in mtt
we ended up with MPI_Wtime() < MPI_Wtick()

I added a new test in ibm/environment
MPI_Wtick() < 1

PR #917, opened 2016-01-25
osc/rdma: fix hang when performing large unaligned gets

hjelmn:
This commit adds code to handle large unaligned gets. There are two
possible code paths for these transactions:

 1) The remote region and local region have the same alignment. In
 this case the get will be broken down into at most three get
 transactions: 1 transaction to get the unaligned start of the region
 (buffered), 1 transaction to get the aligned portion of the region,
 and 1 transaction to get the end of the region.

 2) The remote and local regions do not have the same alignment. This
 should be an uncommon case and is not optimized. In this case a
 buffer is allocated and registered locally to hold the aligned data
 from the remote region. There may be cases where this fails (low
 memory, can't register memory). Those conditions are unlikely and
 will be handled later.

(cherry picked from open-mpi/ompi@45da3114737d876d5c8bab25f61ad45f242d6eba)

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_comm.c
+++ b/ompi/mca/osc/rdma/osc_rdma_comm.c
@@ -15,6 +15,11 @@
 #include "osc_rdma_dynamic.h"
 
 #include "ompi/mca/osc/base/osc_base_obj_convert.h"
+#include "opal/align.h"
+
+static int ompi_osc_rdma_get_contig (ompi_osc_rdma_sync_t *sync, ompi_osc_rdma_peer_t *peer, uint64_t source_address,
+                                     mca_btl_base_registration_handle_t *source_handle, void *target_buffer, size_t size,
+                                     ompi_osc_rdma_request_t *request);
 
 static void ompi_osc_get_data_complete (struct mca_btl_base_module_t *btl, struct mca_btl_base_endpoint_t *endpoint,
                                         void *local_address, mca_btl_base_registration_handle_t *local_handle,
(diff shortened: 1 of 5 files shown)


hjelmn:
:bot:assign: @regrant 
:bot:label:bug
:bot:milestone:v2.0.0

hjelmn:
@regrant Found by the threading tests. When using a non-power of two number of threads I was seeing an infinite loop on XE due to the get alignment requirements. This commit fixes the problem by either buffering large gets into malloced memory (slow) or attempting to break down the get into a large aligned and up to two small unaligned segments. Tested on Cielito.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1255/ for details.

hppritcha:
@hjelmn do you have a test which shows this problem?

hjelmn:
Yes.

jsquyres:
@hjelmn Can you add that test to, say, the ibm test suite?

hjelmn:
Hmm, not sure. It is up to @regrant while the code is still private.

regrant:
@hppritcha @jsquyres @hjelmn The test code is going to be released as part of the Sandia Microbenchmark Suite, as soon as I've got all the local approvals for that, we can add it to the OMPI test suites. We just can't share it outside of the labs quite yet.

jsquyres:
@regrant @hjelmn Ok, gotcha.  If we can get this run during regular MTT runs once its out/available, that would be great.

regrant:
@jsquyres For sure, they were meant to help the community improve Multi-threaded RMA performance in the first place, so sharing them is the ultimate goal here, paperwork is the obstacle.

regrant:
@hjelmn Reviewed, looks good to me :+1:

jsquyres:
@hppritcha Good to go

PR #961, opened 2016-02-11
pmix s2: adjust module structure initialisation to fit pmix.h

plesn:
In v2.x branch the pmix s2 module structure initialisation is out of sync with pmix.h.
This makes launch with "srun --mpi=pmi2" work again.

### DIFF
--- a/opal/mca/pmix/s1/pmix_s1.c
+++ b/opal/mca/pmix/s1/pmix_s1.c
@@ -88,6 +88,8 @@ const opal_pmix_base_module_t opal_pmix_s1_module = {
     NULL,
     NULL,
     NULL,
+    NULL,
+    NULL,
     /* utility APIs */
     NULL,
     opal_pmix_base_register_handler,
--- a/opal/mca/pmix/s2/pmix_s2.c
+++ b/opal/mca/pmix/s2/pmix_s2.c
@@ -95,6 +95,8 @@ const opal_pmix_base_module_t opal_pmix_s2_module = {
     NULL,
     NULL,
     NULL,
+    NULL,
+    NULL,
     /* utility APIs */
     NULL,
     opal_pmix_base_register_handler,


hppritcha:
@plesn  could you explain out of sync better?  is it in sync on master?  we'd rather cherry pick the correct commits from master into v2.x.

hppritcha:
Okay this looks like a big mess.  It appears that these lines were touch by a really big commit on to master open-mpi/ompi@5f446570d

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1329/ for details.

rhc54:
Why would this go into 2.x?? We haven't brought any of that extension over to 2.x that I know about.

plesn:
The number of fields in the initialisation of opal_pmix_s2_module is not the same as in the declaration of opal_pmix_base_module_t, two fields are missing (I wonder how the compiler can let this go…).

Therefore opal_pmix.store_local() ends up calling in fact s2_register_jobid() and makes the pmi2 launch abort.

The code in master does not have this pitfall as it uses designated initializers (I find it cleaner).

rhc54:
We have required C99 since the 1.7 branch, so we might as well update it here. FWIW: the compiler will always NULL-extend the struct, so it is probably the register_errhandler entry that is having the problem as it is at the end of the struct.

plesn:
The two missing fields are in the middle of the struct : it seems to be due to the "server_deregister_{nspace,client}" calls which where added in pmix.h but not in s2 in https://github.com/open-mpi/ompi-release/commit/8abed29c822b24f4c5849efc72819d802a0e714b#diff-78752eed2a7de7b06fd5840b9637b882R777.

On master the struct initialisers are added in 5f446570.

rhc54:
Well, this certainly is the minimal required patch - thanks! Did you by chance look at the s1 and alps components to see if they need to be extended as well?

plesn:
s1 has the same problem, while cray is ok as it uses struct initialisers. So maybe the best solution would be to cherry pick 5f44657 which has both corrections.

rhc54:
Agreed - that should have come over anyway as it has a bunch of memory leak cleanups. I can post it if you are willing to review?

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1331/ for details.

plesn:
I can at least test it :)

hppritcha:
I"m giving this a bug label.

hppritcha:
@rhc54 this PR can be closed owing to #963, correct?

rhc54:
Correct

hppritcha:
i will give this PR a spin on one of the slurm clusters here at LANL - never mind wrong PR.

PR #967, opened 2016-02-15
opal/memory: Move Memory Allocation Hooks usage from openib

igor-ivanov:
See related PR in ompi/trunk https://github.com/open-mpi/ompi/pull/1351

:bot:assign: @jsquyres
:bot:label:bug
:bot:milestone:v2.0.0

ompi-trunk: 
https://github.com/open-mpi/ompi/commit/d9eefefa74ce9ef9217c8ed8aba4cdb819c6fbaa
https://github.com/open-mpi/ompi/commit/477991b5aadd0ccfb4e846ae1f03a0f21b6f8d90
https://github.com/open-mpi/ompi/commit/5c685e23

### DIFF
--- a/opal/mca/memory/linux/memory_linux_component.c
+++ b/opal/mca/memory/linux/memory_linux_component.c
@@ -129,6 +129,14 @@ static void *(*prev_malloc_hook)(size_t, const void *);
  * memalign which will be called through __malloc_hook instead of malloc.
  */
 static void *_opal_memory_linux_malloc_align_hook(size_t sz, const void* caller);
+
+static mca_base_var_enum_value_t align_values[] = {
+    {-1, "disabled"},
+    {0, "0"},
+    {32, "32"},
+    {64, "64"},
+    {0, NULL}
+};
 #endif /* MEMORY_LINUX_MALLOC_ALIGN_ENABLED */
 
 
@@ -137,6 +145,9 @@ static void *_opal_memory_linux_malloc_align_hook(size_t sz, const void* caller)
  */
 static int linux_register(void)
 {
+#if MEMORY_LINUX_MALLOC_ALIGN_ENABLED
+    mca_base_var_enum_t *new_enum;
+#endif
     int ret;
     /* Information only */
     ret = mca_base_component_var_register (&mca_memory_linux_component.super.memoryc_version,
@@ -199,17 +210,19 @@ static int linux_register(void)
     }
 
 #if MEMORY_LINUX_MALLOC_ALIGN_ENABLED
+    (void)mca_base_var_enum_create("memory_linux_memalign", align_values, &new_enum);
     mca_memory_linux_component.use_memalign = -1;
     ret = mca_base_component_var_register(&mca_memory_linux_component.super.memoryc_version,
                                  "memalign",
-                                 "[64 | 32 | 0] - Enable memory alignment for all malloc calls (default: disabled).",
+                                 "[64 | 32 | 0] - Enable memory alignment for all malloc calls.",
                                  MCA_BASE_VAR_TYPE_INT,
-                                 NULL,
+                                 new_enum,
                                  0,
                                  0,
                                  OPAL_INFO_LVL_5,
                                  MCA_BASE_VAR_SCOPE_READONLY,
                                  &mca_memory_linux_component.use_memalign);
+    OBJ_RELEASE(new_enum);
     if (0 > ret) {
         return ret;
     }
@@ -229,16 +242,6 @@ static int linux_register(void)
                                  &mca_memory_linux_component.memalign_threshold);
     if (0 > ret) {
         return ret;
-    }
-
-    if (mca_memory_linux_component.use_memalign != -1
-        && mca_memory_linux_component.use_memalign != 32
-        && mca_memory_linux_component.use_memalign != 64
-        && mca_memory_linux_component.use_memalign != 0){
-        opal_show_help("help-opal-memory-linux.txt", "invalid mca param value",
-                       true, "Wrong memalign parameter value. Allowed values: 64, 32, 0.",
-                       "memory_linux_memalign is reset to 32");
-        mca_memory_linux_component.use_memalign = 32;
     }
 #endif /* MEMORY_LINUX_MALLOC_ALIGN_ENABLED */
 
--- a/opal/mca/memory/linux/memory_linux_component.c
+++ b/opal/mca/memory/linux/memory_linux_component.c
@@ -13,6 +13,8 @@
  * Copyright (c) 2009-2014 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2013-2015 Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -39,12 +41,17 @@
 
 #include "opal_config.h"
 
+#if HAVE_MALLOC_H
+#include <malloc.h>
+#endif
+
 #include "opal/constants.h"
 #include "opal/mca/base/mca_base_var.h"
 #include "opal/mca/memory/memory.h"
 #include "opal/mca/memory/base/empty.h"
 #include "opal/memoryhooks/memory.h"
 #include "opal/util/output.h"
+#include "opal/util/show_help.h"
 
 #include "opal/mca/memory/linux/memory_linux.h"
 #undef opal_memory_changed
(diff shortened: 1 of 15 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1339/ for details.

jsquyres:
Per https://github.com/open-mpi/ompi/pull/1351#discussion-diff-52908786, this PR contains an abstraction violation that needs to be fixed.  :-1:

igor-ivanov:
@jsquyres could you please propose correct way.

jsquyres:
Typically, you have to do these kinds of things via indirection (e.g., via the framework base).  This will also allow you to get rid of the `#if` in the BTL code.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1345/ for details.

igor-ivanov:
@jsquyres could you check this update

opal/mca/btl/base/btl_base_select.c
```diff
@@ -107,6 +122,16 @@ int mca_btl_base_select(bool enable_progress_threads,
                                 "select: no init function; ignoring component %s",
                                 component->btl_version.mca_component_name);
         } else {
+#if MEMORY_LINUX_MALLOC_ALIGN_ENABLED
+            if (strcmp(component->btl_version.mca_component_name, "openib") == 0) {
```
jsquyres: You can't put a check for a specific component in a framework base.

jsquyres:
@igor-ivanov Can you please revert the master commit?  The abstraction violation is causing problems for others: http://www.open-mpi.org/community/lists/devel/2016/02/18598.php.  Thanks.

opal/mca/memory/linux/memory_linux_component.c
```diff
@@ -162,6 +187,53 @@ static int linux_register(void)
...
                                             &opal_memory_linux_disable);
+    if (0 > ret) {
+        return ret;
+    }
+
+#if MEMORY_LINUX_MALLOC_ALIGN_ENABLED
+    mca_memory_linux_component.use_memalign = -1;
+    ret = mca_base_component_var_register(&mca_memory_linux_component.super.memoryc_version,
+                                 "memalign",
+                                 "[64 | 32 | 0] - Enable memory alignment for all malloc calls (default: disabled).",
```
jsquyres: Is there a reason not to use an MCA var enumeration here?

opal/mca/btl/base/btl_base_select.c
jsquyres: Ah -- I see the problem: I think I wasn't clear in my comment to you before about doing this "through the framework base."  Sorry.  It was clear in my head! :wink: Here's what I meant:

I was actually referring to the `opal/mca/memory` base (not the BTL base).  I.e., make an OPAL memory framework function to set a memalign hook.  This function would be located in `opal/mca/memory/base`, and openib (and oshmem, and... whoever else needs it) can just call it directly, without needing to directly include `opal/mca/memory/linux/something.h` or directly invoke a function in `opal/mca/memory/linux`.  Instead, the memory/base function will lookup and see if the selected memory component supports setting a memalign hook, and if so, set it (e.g., perhaps by calling a new member function on the memory module structure).  The linux memory hook component will support this functionality; I don't know if the solaris one will (i.e., it needs to be safe for the solaris memory component to _not_ support this functionality).

Does that make more sense?

opal/mca/memory/linux/memory_linux_component.c
igor-ivanov: seems @ggouaillardet solved issue in master. Thanks.

only reason is to minimize changes with original code. Moreover using enumeration requires additional variable.

hppritcha:
Any idea why Jenkins/Travis did not catch this?

opal/mca/btl/base/btl_base_select.c
igor-ivanov: @jsquyres thank you for sharing details. Actually my initial local version did something you described. Reason I did not push one was in necessity to use #ifdef inside memory/base that relates memory/linux component only. It seems that solaris does not use hooks.

jsquyres: Fair enough.  Perhaps this could be exposed as a function pointer on the module -- hence, the solaris module can provide a NULL pointer, and the linux one can provide a non-NULL pointer.

opal/mca/memory/linux/memory_linux_component.c
jsquyres: True, but this is exactly the case that enumerated MCA vars were intended for.

jsquyres:
I think @ggouaillardet did a workaround on master that's probably good enough for the moment, but the final result should probably be synchronized between master and v2.x.

igor-ivanov:
@jsquyres Could you advice.
There is memory/linux component ant it looks as suitable for keeping malloc hook operation. Now memory/linux component can be disabled in case --with-memory-manager=no is selected but we would like to have hook operation under Linux despite this (original openib is able to support this capability in both situations). So impementing this stuff inside memory/linux does not help completely. 
I see a moving mca variables and hook operations on memory/base layer as a way that could solve this issue (note: there is 'empty' component inside 'base' layer). configure check if memory hook exists should be moved to memory top layer also.
Do you see any abstraction violation issues in this approach. What is your opinion.

hppritcha:
Does this also need to go in to 1.10.X?

ggouaillardet:
once open-mpi/ompi#1376 is merged, my last three commits can be cherry-picked into this PR and it should be good to go

igor-ivanov:
One point here.
Original code (before https://github.com/open-mpi/ompi/pull/1351) was able to use aligned malloc in case no memory-manager is enabled (actually memory/empty is activated that is hidden in memory/base). This ability was lost after the PR (my apologies). See https://github.com/open-mpi/ompi-release/pull/967#issuecomment-185250851.
I have no idea how to solve this w/o breaking abstraction. There is 'linux' component that is suitable for this but it can be disabled. I can not add mostly linux specific feature in 'empty' component too.

ggouaillardet:
if memory alignment is not specific to the Linux component, then it should be moved to base.
an other option is to write an other minimal component that does alignment only.

igor-ivanov:
minimal component can be disabled as --memory-manager=none up layer option. So it seems not good way. As for the second option that I asked @jsquyres to check.

ggouaillardet:
well, if the **only** role of the minimal memory manager is to align memory, then you cannot have memory aligned with --memory-manager=none
that sounds like a feature to me...

igor-ivanov:
After looking at old openib code more I realized that malloc hook did not work w/o memory manager. Setting such hook was controled by opal_mem_hooks_support_level(). Only memory/linux and memory/malloc_solaris set hooks support level calling opal_mem_hooks_set_support().
So changes from @ggouaillardet should be enough.

mike-dubman:
@hppritcha - yes, we would like to port it into v1.10.x once ready.

mike-dubman:
:+1:

ggouaillardet:
@miked-mellanox is this pr ready to be merged ?

I would have expected it is updated with commits cherry-picked from master, especially the one in open-mpi/ompi#1351 (and I suspect some previous commits should be cherry-picked too)

mike-dubman:
@igor-ivanov - could you please comment? thanks

jsquyres:
Deferring until @ggouaillardet's questions are answered... :-1:

lanl-ompi:
Test FAILed.

mellanox-github:
Test FAILed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1381/ for details.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1382/ for details.

mike-dubman:
:+1:

mike-dubman:
:-1:

mike-dubman:
:+1:

mike-dubman:
coool :)

igor-ivanov:
@ggouaillardet could you check if you see any issue

jsquyres:
@igor-ivanov I think this looks good; nice job.  I'd still like to see the MCA enumeration, though.

@ggouaillardet Does it look good to you?

ggouaillardet:
@jsquyres so far so good, i made open-mpi/ompi#1420 for the MCA enumeration

hppritcha:
@jsquyres i pronounce ready to go.

jsquyres:
@hppritcha I'd like to see a v2.0.0 PR corresponding to open-mpi/ompi#1420 first.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1397/ for details.

jsquyres:
:+1:

PR #997, opened 2016-03-02
Repair the application suicide code for 2.0 - requires custom patch.

rhc54:
The link from a PMIx comm failure back to the ORTE errmgr had been broken. Reconnect it so that the app's errmgr can abort upon loss of connection to the local daemon

@jsquyres The problem was that the call to PMIx_Register_errhandler wasn't implemented, and so we never connected to the PMIx errhandler to the ORTE errmgr even though we were registering it. This implements the registration code so the connection gets made.

Fixes open-mpi/ompi#1418

### DIFF
--- a/opal/mca/pmix/base/pmix_base_fns.c
+++ b/opal/mca/pmix/base/pmix_base_fns.c
@@ -2,7 +2,7 @@
 /*
  * Copyright (c) 2012-2015 Los Alamos National Security, LLC.  All rights
  *                         reserved.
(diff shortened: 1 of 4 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1393/ for details.

jsquyres:
@gpaulsen Github pro tip: it is usually better to comment in the "Files changed" view on the PR itself.

jsquyres:
@gpaulsen Gave this a :+1:

PR #1001, opened 2016-03-04
pmi2 disqualified in singloton startup

plesn:
In openmpi <=1.10, there a bug when launching with PMI2 in singloton mode :

```
$ ./mpi_hello
[...]
 PMI2_Job_GetId failed failed
  --> Returned value (null) (14) instead of ORTE_SUCCESS
```

A possible correction is to check at pmi2 init if we get a job id.

### DIFF
--- a/opal/mca/common/pmi/common_pmi.c
+++ b/opal/mca/common/pmi/common_pmi.c
@@ -46,7 +46,6 @@ bool mca_common_pmi_init (void) {
         rank = -1;
         appnum = -1;
 
-
         /* if we can't startup PMI, we can't be used */
         if (PMI2_Initialized ()) {
             return true;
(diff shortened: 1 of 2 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1398/ for details.

plesn:
On slurm 14.11 the first commit (9c83f9d) still leaves a scary warning
saying that PMI2 returned bad values.

Checking errors for PMI2 and handling the warning are in fact 2
separate things:
1. depending on slurm versions, PMI2 fails differently:
   - in 14.11, PMI2_Init() succeeds but gives incorrect values
   - in 15.08, rank/size are correct, but we are in degraded mode:
     many functions requiring a server like PMI2_Job_GetId() fail.
2. PMI2 can fail to contact its server for at least 2 different
   reasons:
   - no slurm flag `--mpi=pmi2` given. _This needs a warning_
   - singleton : no server when no srun. _This needs no warning_
3. proper singleton check requires to check if there is the Slurm or
   Cray launcher.

So I propose to check PMI errors and warn (commit bb65c3f) and just
quiet the warning when there is no srun (even if it is a bit hacky).

mellanox-github:
Test FAILed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1401/ for details.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1402/ for details.

rhc54:
This is a puzzler as the root cause of the problem is that SLURM's PMI2 library incorrectly returns "success" even when it "fails". I gather they now give some correct values, but the server is still messed up. Sigh.

I'm not wild about this fix, but I honestly can't think of anything better. :+1:

PR #1004, opened 2016-03-08
coll/libnbc: do not handle MPI_IN_PLACE in neighborhood collectives

ggouaillardet:
This is a one off and a quick fix, and it unlikely works with complex types
(e.g. negative lower bound).
libnbc considers MPI_Ineighbor_alltoallw is in place when send and recv buffers are identical,
and even if they do not overlap because of distinct displacements.

Thanks Jun Kudo for the bug report.

### DIFF
--- a/ompi/mca/coll/libnbc/nbc_ineighbor_allgather.c
+++ b/ompi/mca/coll/libnbc/nbc_ineighbor_allgather.c
@@ -1,9 +1,11 @@
 /*
- * Copyright (c) 2006 The Trustees of Indiana University and Indiana
- *                    University Research and Technology
- *                    Corporation.  All rights reserved.
- * Copyright (c) 2006 The Technical University of Chemnitz. All
- *                    rights reserved.
+ * Copyright (c) 2006      The Trustees of Indiana University and Indiana
+ *                         University Research and Technology
+ *                         Corporation.  All rights reserved.
+ * Copyright (c) 2006      The Technical University of Chemnitz. All
+ *                         rights reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  *
  * Author(s): Torsten Hoefler <htor@cs.indiana.edu>
  *
@@ -56,13 +58,10 @@ int ompi_coll_libnbc_ineighbor_allgather(void *sbuf, int scount, MPI_Datatype st
   res = MPI_Type_extent(rtype, &rcvext);
   if (MPI_SUCCESS != res) { printf("MPI Error in MPI_Type_extent() (%i)\n", res); return res; }
 
-  char inplace;
   NBC_Schedule *schedule;
 #ifdef NBC_CACHE_SCHEDULE
   NBC_Ineighbor_allgather_args *args, *found, search;
 #endif
-
-  NBC_IN_PLACE(sbuf, rbuf, inplace);
 
   handle->tmpbuf=NULL;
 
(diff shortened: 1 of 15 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1403/ for details.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1404/ for details.

rhc54:
@ggouaillardet can you please request a reviewer?

ggouaillardet:
:bot:assign: @bosilca 

see my comments at http://www.open-mpi.org/community/lists/users/2016/03/28656.php

bosilca:
@ggouaillardet I don't think this patch is correct. The MPI_INEIGHBOR_ALLTOALLW function (defined in MPI 3.1 page 328) identifies rdispl as the

> integer array (of length indegree). Entry i specifies the displacement in bytes (relative to recvbuf) at which to place the incoming data from neighbor i (array of integers)

Thus, your patch that does compute the receive buffer for each MPI_Recv using the recvtypes and recvcounts for the particular peer does not respect the MPI standard. I think the correct receive buffer for each planned receive should be `NBC_Sched_recv((char*)recvbuf+rdisps[i]...`

ggouaillardet:
@bosilca 
thanks for the review
i fixed the `inplace` case only, and kept the change as minimal as possible
since we are `inplace`, data is sent from `sendbuf` and received into a temporary buffer, and at the end it is copied into `recvbuf`
if i use `rdisps` to compute offset into the temporary buffer, that means i might have to allocate a bigger buffer and that can be simply avoided.

as i wrote in the ML, i had several interrogations
- is it valid to call `MPI_Ineighbor_alltoallw` with `sendbuf`` ==`recvbuf``` ?
- `NBC_IN_PLACE` is very primitive and not robust
  if `sendbuf == recvbuf` but `sdpisps`` and`rdisps`are built so data to be sent and data to be received do not overlap (see user provided test case) then`inplace`could/should be set to false, assuming this is valid per the mpi standard
  if`sendbuf != recvbuf`but`sdisps`and`rdisps`are built so data to be sent and data to be received do overlap, then`inplace``` should be true, assuming this is a valid use case per the mpi standard

bottom line, if the user test case is valid from the MPI standard, then i think this commit is helping.
master does handle things differently, so i initially did not plan to invest much efforts on this.

bosilca:
@ggouaillardet I am sorry I missed both the fact that you were in the INPLACE branch and your email on the mailing list. Let me try to quickly remedy to this.
1. If the data overlap between any tuple of (sendbuf + sdispl[i], stype[i], scount[i]) and (recvbuf + rdispl[i], rtype[i], rcount[i]) the call would be incorrect.
2. If the user does craft the displacements in a way to avoid any overlap, then technically we are not in an inplace code.

Thus, the inplace code should be entirely stripped out. I guess @hjelmn reached the same conclusion when he removed the inplace support from master for all the neighborhood collectives (d42e0968b15) (and here I am not talking about the MPI_IN_PLACE which has no meaning for the neighborhood collective).

ggouaillardet:
@bosilca i quickly checked the standard and i reached the same conclusion
- `MPI_IN_PLACE` has no meaning for neighborhood collectives
- it is the user responsability to ensure there is no overlapping between send and receive buffers
  (ompi does not check that, we could do that and return `MPI_ERR_BUFFER`)
  so i will remove the `inplace` from the neighborhood collectives from now

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1416/ for details.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1417/ for details.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1418/ for details.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1422/ for details.

jsquyres:
@bosilca Can you have a look at this?

bosilca:
The patch removes all support for MPI_INPLACE from neighborhood collective, which put us in compliance with the MPI standard. :+1:

jsquyres:
@hppritcha Good to go.

rhc54:
well, I approve it too :smile:

PR #1051, opened 2016-03-28
usnic: remove need for hwloc verbs helper

jsquyres:
Haven't needed this for a while, but it got left in by accident.

(cherry picked from commit open-mpi/ompi@4a3c986a80ed1c26b4f5ab62197b222272a93d64)

This actually does fix a problem when you try to build with libfabric support with an external hwloc that doesn't have verbs support built in.  It's a minor corner case, though (unless @rhc54 disagrees -- he's the one who found this issue).

### DIFF
--- a/opal/mca/btl/usnic/btl_usnic_hwloc.c
+++ b/opal/mca/btl/usnic/btl_usnic_hwloc.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2015 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2013-2016 Cisco Systems, Inc.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1479/ for details.

rhc54:
FWIW: I ran into it when trying to build with our internal hwloc.

:+1:

jsquyres:
@rhc54 Do you need this in 2.0.0, or is it ok if it waits until 2.0.1?

rhc54:
it can wait

gpaulsen:
yes please.  Can wait for 2.0.1.

rhc54:
I didn't recall this being over here - it can wait for me as I'll just disable the usnic btl. Not sure if it is a "real-world" issue - what happens if someone has an installed hwloc in their path? Won't it cause the same problem for them?

jsquyres:
I think the risk is low to pull this PR back from v2.0.1 to v2.0.0.

On the one hand, I think you're running into a corner case: I think the vast majority of people use the embedded hwloc, so they don't run into this issue.  OTOH, some people may run into it who are using OPAL outside of OMPI (like you are).

@hppritcha Got an opinion?

hppritcha:
I'm okay with merging this.

PR #1118, opened 2016-04-30
Fix the PMIx external library config so that it checks both lib and lib64 locations

rhc54:
Thanks to Orion Poplawski for pointing it out

The patch for master will likely be different as we are dealing with multiple versions, so this is just for 2.x for now.


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1578/ for details.

rhc54:
Fixes https://github.com/open-mpi/ompi/issues/1609

ggouaillardet:
@rhc54 i made open-mpi/ompi#1612, using `OPAL_CHECK_PACKAGE` macro makes it much easier to me

jsquyres:
#1612 does seem quite a bit simpler, and uses the macro as intended.

@rhc54 Is there a reason not to use `OPAL_CHECK_PACKAGE`?

ggouaillardet:
fwiw, pmix used to use more include dirs than necessary, and hence `OPAL_CHECK_PACKAGE` was not a good fit. I reduced that to one include directory in order to use the macro.

rhc54:
@ggouaillardet Did you read the comment explaining why we cannot use check_package??

``` c
            # cannot use check_package because there are
            # external dependencies to make the headers
            # build, so just check for presence of header
            # and library files - these checks will error
            # out if the files aren't found, which is okay
            # as we are only executing here if the user
            # specified external pmix
```

ggouaillardet:
I probably read the comment, but could not find such dependencies
(e.g. with my changes, it works just fine, with lib pmix,so in /lib or /lib64)

rhc54:
Yes - IF and only if the external dependencies are all in standard locations! We tried this approach and had to reject it because it fails if things are in non-standard locations since you cannot know what PMIx linked itself against, and where those packages are located.

jsquyres:
@ggouaillardet I think @rhc54 is saying that `pmix.h` includes other headers.  If those headers are not in a standard location, then the include test fails (and it gets complicated/undesirable to try to list all the `-I` flags for the PMIx header dependencies just for this test).

jsquyres:
It's ugly, but it works. 👍

ggouaillardet:
OpenMPI configure detects libevent and hwloc **before** pmix, and hence CPPFLAGS and LDFLAGS are already set when external pmix is checked.
am I right ?
if yes, what can possibly go wrong ?
or let me put it this way, if OPAL_CHECK_PACKAGE fails but current mechanism works, can we ensure 100% everything will work just fine at runtime ?
fwiw, in my vm , libevent is in a standard location, but pmix is in a non standard one

rhc54:
@jsquyres and I just spent a bunch of time on the phone over this, and I think we concur that using OPAL_CHECK_PACKAGE should be okay for now. The reason we can do it is that check_package only does AC_TRY_LINK - it doesn't actually try to run anything, and thus the pmix external linkages don't matter so long as the header for those external links aren't directly referenced in the pmix headers.

So I will close this one out and we can use yours. Thx!

PR #1121, opened 2016-05-02
Allow NULL arrays for empty datatypes.

jsquyres:
When building an empty datatype (aka. size = 0) because the count of
included datatypes is 0, be less strict on what the arguments are
(allow NULL pointers).

(cherry picked from commit open-mpi/ompi@6e6ed62a3c3a58b9352b7ac28418a1504c1baa68)

@bosilca Please review.  Straightforward cherry-pick from master.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -57,6 +57,8 @@ included in the vX.Y.Z section and be denoted as:
 1.10.3 - DATE
 ------
 
+- Fix zero-length datatypes.  Thanks to Wei-keng Liao for reporting
+  the issue.
 - Minor manpage cleanups
 - Implement atomic support in OSHMEM/UCX
 - Fix support of MPI_COMBINER_RESIZED. Thanks to James Ramsey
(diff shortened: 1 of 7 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1582/ for details.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1584/ for details.

bosilca:
This is the cherry-pick from master. This particular commit looks OK but I noticed something strange while looking at the code.

In 1.10 the MPI_Type_hindexed explicitly calls MPI_Type_create_hindexed. I was under the impression that we were trying to never call an MPI function from inside the library.

jsquyres:
@bosilca Good catch.  Looks like we have fixed this kind of thing for the v2.0.0 release -- there's a handful of MPI functions that turn around and call other MPI functions (E.g., MPI_ERRHANDLER_GET calls MPI_COMM_GET_ERRHANDLER).  In master/v2.x, they actually call the PMPI equivalents, so that we at least don't trigger tools improperly.

But other than that, it sounds like you're cool with the cherry pick: 👍

PR #1147, opened 2016-05-09
Pr/v1.10/further updates to so versions

jsquyres:
Also bump to 1.10.3rc2

@rhc54 Please review.  If good, let me know and I'll remove the comments before merging.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -24,7 +24,7 @@ release=3
 # requirement is that it must be entirely printable ASCII characters
 # and have no white space.
 
-greek=rc1
+greek=rc2
 
 # If repo_rev is empty, then the repository version number will be
 # obtained during "make dist" via the "git describe --tags --always"
--- a/VERSION
+++ b/VERSION
@@ -86,10 +86,10 @@ libmpi_so_version=12:3:0
 libmpi_cxx_so_version=2:3:1
 libmpi_mpifh_so_version=12:1:0
 libmpi_usempi_tkr_so_version=6:0:1
-libmpi_usempi_ignore_tkr_so_version=7:0:1
+libmpi_usempi_ignore_tkr_so_version=8:0:2
 libmpi_usempif08_so_version=12:1:1
 libopen_rte_so_version=12:3:0
-libopen_pal_so_version=13:2:0
+libopen_pal_so_version=13:3:0
 libmpi_java_so_version=3:0:2
 liboshmem_so_version=9:1:1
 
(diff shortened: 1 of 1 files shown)


mellanox-github:
Test FAILed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1627/ for details.

jsquyres:
@jladd-mlnx @Di0gen This looks like a local failure on the Mellanox jenkins -- can you fix?

```
11:29:50 --- MCA component pml:ucx (m4 configuration macro)
11:29:50 checking for MCA component pml:ucx compile mode... dso
11:29:50 checking --with-ucx value... sanity check ok (/hpc/local/benchmarks/hpc-stack-gcc/install/ucx)
11:29:50 checking --with-ucx-libdir value... simple ok (unspecified)
11:29:50 checking ucp/api/ucp.h usability... yes
11:29:50 checking ucp/api/ucp.h presence... yes
11:29:50 checking for ucp/api/ucp.h... yes
11:29:50 checking for library containing ucp_cleanup... no
11:29:50 checking for UCX version compatibility... yes
11:29:50 configure: error: UCX support requested but not found.  Aborting
```

VERSION
```diff
@@ -83,15 +83,23 @@ date="Unreleased developer copy"
...
 libmpi_so_version=12:3:0
+# was 12:2:0
 libmpi_cxx_so_version=2:3:1
+# was 2:3:1
 libmpi_mpifh_so_version=12:1:0
+# was 12:0:0
 libmpi_usempi_tkr_so_version=6:0:1
-libmpi_usempi_ignore_tkr_so_version=7:0:1
+libmpi_usempi_ignore_tkr_so_version=8:0:2
+# was 7:0:1
```
rhc54: If we are upping the primary number, why is the release number not 0?

jsquyres: These are not major.minor.release numbers -- these are current:revision:age triples, as described in the Libtool docs -- see the rules listed in section 7.3: https://www.gnu.org/software/libtool/manual/libtool.html#Updating-version-info

rhc54: I understand that - my understanding from the libtool docs is that a change to the APIs requires a change to the "current" component, and that must in turn reset "revision" and "age" to 0.

jsquyres: Ah, sorry.

What happened with this triple is that we added some APIs, so we bumped up `current`.  But we didn't make any backwards incompatibilities, so I bumped up `age`, too (so that both 7:0:1 and 8:0:2 both support 6:0:0).

rhc54: Ugh...okay, I guess. Not the way I interpreted the rules, but I can accept it.

jsquyres:
Per discussion here and in a few subsequent emails, 👍  from @rhc54

PR #1166, opened 2016-05-17
When direct launching applications, we must allow the MPI layer to progress during RTE-level barriers.

rhc54:
Neither SLURM nor Cray provide non-blocking fence functions, so push those calls into a separate event thread (use the OPAL async thread for this purpose so we don't create another one) and let the MPI thread sping in wait_for_completion. This also restores the "lazy" completion during MPI_Finalize to minimize cpu utilization.

Update external as well

Revise the change: we still need the MPI_Barrier in MPI_Finalize when we use a blocking fence, but do use the "lazy" wait for completion. Replace the direct logic in MPI_Init with a cleaner macro

(cherry picked from commit open-mpi/ompi@01ba861f2ab96c0562464d16cb1b28c40d408a94)

Fix registration of error handlers thru the pmix120 component. A thread-shift operation was hanging on the sync_event_base, which made it dependent on someone calling opal_progress. Unfortunately, a process in "sleep" or spinning outside the MPI library won't do that, and so we never complete errhandler registration.

(cherry picked from commit open-mpi/ompi@4a55fba4147af996813e8e1d8df62b64b94efca3)

### DIFF
--- a/ompi/mca/rte/rte.h
+++ b/ompi/mca/rte/rte.h
@@ -207,27 +207,27 @@ OMPI_DECLSPEC extern mca_base_framework_t ompi_rte_base_framework;
  * progress while waiting, so we loop over opal_progress, letting
  * the RTE progress thread move the RTE along
  */
-#define OMPI_WAIT_FOR_COMPLETION(flg)                                   \
-    do {                                                                \
-        opal_output_verbose(1, ompi_rte_base_framework.framework_output, \
-                            "%s waiting on RTE event at %s:%d",         \
-                            OMPI_NAME_PRINT(OMPI_PROC_MY_NAME),         \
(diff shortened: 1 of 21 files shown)


rhc54:
@matcabral Can you please verify as well?

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1671/ for details.

matcabral:
@rhc54 Tested on a single node with SLURM and confirmed the patch fixes the hang issue seen on open-mpi/ompi#1559 on branch v.2.x

matcabral:
fixes open-mpi/ompi#1559

hppritcha:
works for me

hppritcha:
@jsquyres ready to go

PR #1181, opened 2016-05-23
Update the Linux timer per the master and George's patch on the devel mailing list.

rhc54:
@bosilca Can you please review? Your patch on the devel mailing list did not apply cleanly to 1.10, so I tried my best to capture things from master.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -841,7 +841,10 @@ OPAL_SEARCH_LIBS_CORE([dirname], [gen])
 # Darwin doesn't need -lm, as it's a symlink to libSystem.dylib
 OPAL_SEARCH_LIBS_CORE([ceil], [m])
 
-AC_CHECK_FUNCS([asprintf snprintf vasprintf vsnprintf openpty isatty getpwuid fork waitpid execve pipe ptsname setsid mmap tcgetpgrp posix_memalign strsignal sysconf syslog vsyslog regcmp regexec regfree _NSGetEnviron socketpair strncpy_s usleep mkfifo dbopen dbm_open statfs statvfs setpgid  setenv])
+# -lrt might be needed for clock_gettime
+OPAL_SEARCH_LIBS_CORE([clock_gettime], [rt])
+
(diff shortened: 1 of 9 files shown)


mellanox-github:
Test FAILed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1693/ for details.

bosilca:
This is indeed an old patch open-mpi/ompi@00300f46. 👍

rhc54:
Not sure of the source of the issue with pthread_atfork on Mellanox - not seeing an issue anywhere else. I will take a look, but don't have much time to spend on it. Suggest that perhaps Mellanox diagnose why their environments are missing it.

rhc54:
Built it again just fine on CentOS7, so I suspect indeed that this is an issue of Mellanox not having a pthread library installed. Note also that this was in last night's tarball, and that no MTT failures were reported this morning.

jladd-mlnx:
@rhc54 Negative, Ghostrider. We have pthread installed. 
$ls -l /usr/include/pthread.h
-rw-r--r-- 1 root root 40491 Nov  5  2013 /usr/include/pthread.h
$ls -l /usr/lib64/libpthread.so
-rw-r--r-- 1 root root 222 Nov  5  2013 /usr/lib64/libpthread.so

Git bisect squarely points to this PR.

jsquyres:
@jladd-mlnx See open-mpi/ompi#1699.  Do you have someone who can look into this?

jladd-mlnx:
@jsquyres I do not. We are trying to get an HPC-X release done this week.

jladd-mlnx:
@rhc54 Can you revert this? It's causing all kinds of problems with our QA and we're in a critical release phase with both HPC-X and MOFED.

bosilca:
I don't know why -lpthread would be required for this patch. It does however need -lrt.

jsquyres:
@bosilca See the detailed analysis on https://github.com/open-mpi/ompi/issues/1699

PR #1209, opened 2016-06-02
While we weren't really hanging, it was taking a very, very long time…

rhc54:
… to abort processes due to the wait time introduced into the sequential kill operation. Provide a delay in a different way so we can exit faster.


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1732/ for details.

PR #1214, opened 2016-06-06
coll/basic: fix mca_coll_basic_bcast_log_intra()

ggouaillardet:
The log basic bcast was completely broken. The rank 0 gets the
hibit set to -1, so it always returned an error.

(cherry picked from commit open-mpi/ompi@c976559877d8b5bf23e52c234d011771ae44a1d2)

### DIFF
--- a/ompi/mca/coll/basic/coll_basic_bcast.c
+++ b/ompi/mca/coll/basic/coll_basic_bcast.c
@@ -9,7 +9,9 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2015      Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


ggouaillardet:
:bot:label:bug
:bot:label:blocker
:bot:milestone:v2.0.0
:bot:assign: @bosilca 

This is a similar fix to open-mpi/ompi@9376b0340b59b702aaa381083063c74287e6f3fb

the regression was introduced when trying to silence the following coverity warning
`1 << hibit` with a negative `hibit`

_we_ know that `hibit >=0` if `rank > 0` but coverity does/did not, hence the warning.

i added an assert to silence the warning.
that being said, i have a recollection of coverity **not** using the assert in order to silence these kind of false positive. if a warning is still being issued, then i will try something more coverity specific

``` patch
diff --git a/ompi/mca/coll/basic/coll_basic_bcast.c b/ompi/mca/coll/basic/coll_basic_bcast.c
index 9dbbb9a..87b7427 100644
--- a/ompi/mca/coll/basic/coll_basic_bcast.c
+++ b/ompi/mca/coll/basic/coll_basic_bcast.c
@@ -68,7 +68,7 @@ mca_coll_basic_bcast_log_intra(void *buff, int count,
     /* Receive data from parent in the tree. */

     if (vrank > 0) {
-        assert(hibit >= 0);
+        /* coverity[BAD_SHIFT] */
         peer = ((vrank & ~(1 << hibit)) + root) % size;

         err = MCA_PML_CALL(recv(buff, count, datatype, peer,
```

based on the coverity result, i will add an assert or coverity directive in `mca_coll_basic_bcast_log_intra`` as well

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1742/ for details.

bosilca:
👍

rhc54:
@ggouaillardet Remember that Coverity may not be doing debug builds, and so the assert may be compiled out and won't be considered as a sufficient protection

ggouaillardet:
@rhc54 that is a very good point, and that would explain why i could find several posts on google that suggests adding an `assert()` in order to get rid of false positive.

could/should we change the coverity setup and have it use debug builds instead ?
iirc, coverity runs on an Open MPI managed server and pushes the results to the web interface.
last coverity scan was on 4/17, so i guess something is not currently working.

jsquyres:
Coverity problems should now be resolved -- there was a problem on their end.  Looks like a new build is processing through their analyzer right now...

PR #1247, opened 2016-06-28
OSHMEM: Removing erroneous initialization check

shamisp:
Since the introduction of the on-demand proc allocation
the check become erroneous and irrelevant.
Moreover, it completely breaks OpenSHMEM support in OMPI.

Signed-off-by: Pavel Shamis (Pasha) pasharesearch@gmail.com

### DIFF
--- a/oshmem/proc/proc.c
+++ b/oshmem/proc/proc.c
@@ -4,6 +4,7 @@
  * Copyright (c) 2014-2015 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2016      ARM, Inc. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


shamisp:
This is a followup commit on the bugfix open-mpi/ompi#1817

jsquyres:
@shamisp This breaks with UCX, right?  I think we determined earlier today that it doesn't break OSHMEM with MXM because MXM disables on-demand addprocs (which is why we hadn't seen this before).

@hppritcha We discussed this earlier today.  If this passes CI (it might still run into the RDMACM problem at Mellanox Jenkins), I think we should merge so that we can get an MTT night in with it.

shamisp:
I don't have MXM to try but I suspect it does. It breaks yoda for sure. I have not seen any special disable code in OSHMEM code. I only saw the code in the OMPI codebase which is not shared with OSHMEM.

mellanox-github:
Test FAILed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1817/ for details.

hppritcha:
I"m good with merging this.  I'm still confused as to how the mlnx jenkins oshmem tests were working though.

PR #1275, opened 2016-07-27
Fix the corner-case of a process running connect-accept on itself.

rhc54:
Fixes #1905


rhc54:
@hjelmn I think we need a version of this in master too - let me know what you think.

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/1977/ for details.

rhc54:
@hjelmn @ggouaillardet 

I do not think the comparison check here is correct. If the user passed `MPI_COMM_WORLD` instead of `MPI_COMM_SELF`, the comparison would fail and we would still hang.

What we really need to check is if the two procs involved in the exchange (i.e., the two "leaders" of the connect/accept operation) are the same process. If that is the case, then we don't do the exchange.

I'm not sure how to perform that test in the required locations. Can you advise?

PR #1324, opened 2016-08-18
update version file to rev 1 for 2.0.1

hppritcha:
@jsquyres 
double-checked with a dist tarball

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -23,7 +23,7 @@ release=1
 # requirement is that it must be entirely printable ASCII characters
 # and have no white space.
 
-greek=a1
+greek=
 
 # If repo_rev is empty, then the repository version number will be
 # obtained during "make dist" via the "git describe --tags --always"
@@ -81,16 +81,16 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=20:0:0
-libmpi_cxx_so_version=20:0:0
-libmpi_mpifh_so_version=20:0:0
-libmpi_usempi_tkr_so_version=20:0:0
-libmpi_usempi_ignore_tkr_so_version=20:0:0
-libmpi_usempif08_so_version=20:0:0
-libopen_rte_so_version=20:0:0
-libopen_pal_so_version=20:0:0
-libmpi_java_so_version=20:0:0
-liboshmem_so_version=20:0:0
+libmpi_so_version=20:1:0
+libmpi_cxx_so_version=20:1:0
+libmpi_mpifh_so_version=20:1:0
+libmpi_usempi_tkr_so_version=20:1:0
+libmpi_usempi_ignore_tkr_so_version=20:1:0
+libmpi_usempif08_so_version=20:1:0
+libopen_rte_so_version=20:1:0
+libopen_pal_so_version=20:1:0
+libmpi_java_so_version=20:1:0
+liboshmem_so_version=20:1:0
 
 # "Common" components install standalone libraries that are run-time
 # linked by one or more components.  So they need to be versioned as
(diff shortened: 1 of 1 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/2098/ for details.

jsquyres:
@hppritcha I do not think that this is right.
1. You can leave the `greek` value set to `a1`.
   1. We leave it set to a non-blank value in the committed `VERSION` so that if any random developer runs `make dist` from their personal git clone, it definitely will have the greek value set, and therefore will not be confused with a real / production / officially released tarball.
   2. The `make_dist_tarball` script that we use to make real / production / official release tarballs will locally strip out the greek value during the build process (but will not git commit / push it back to the central repo).
2. [These are the rules we follow for updating the Libtool `c:r:a` shared library version numbers at each release](https://ftp.gnu.org/pub/old-gnu/Manuals/libtool-1.4.2/html_node/libtool_35.html#SEC35).  Specifically: updating the Libtool `c:r:a` numbers for shared libraries is not supposed to track package versions.  I.e., you don't just update the `r` number to "1" just because we are releasing v2.0.1.

👎

hppritcha:
Okay.  Then do we need to change any version numbers for the 2.0.1 release?  I'd been thinking of rule 3 in the referenced link.

jsquyres:
I normally manually examine the git logs and see exactly what files changed since the last release.  _Most_ changes are in components, and therefore don't affect the `.so` version numbers.

If they're in a core library (e.g., libmpi), I go look at that commit and apply the Libtool rules.  E.g., increment both `c` and `a` if an interface was added, ...etc.

I just did this exercise on the v2.x branch by examining all the commits since v2.0.0.  Here's what I came up with (I added a comment -- which does not need to be in the final commit -- explaining why the shared library version number changed the way it did):

``` patch
diff --git a/VERSION b/VERSION
index 5763ae7..94a76dc 100644
--- a/VERSION
+++ b/VERSION
@@ -81,16 +81,20 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.

-libmpi_so_version=20:0:0
+# libmpi changed
+libmpi_so_version=20:1:0
 libmpi_cxx_so_version=20:0:0
 libmpi_mpifh_so_version=20:0:0
 libmpi_usempi_tkr_so_version=20:0:0
 libmpi_usempi_ignore_tkr_so_version=20:0:0
 libmpi_usempif08_so_version=20:0:0
 libopen_rte_so_version=20:0:0
-libopen_pal_so_version=20:0:0
-libmpi_java_so_version=20:0:0
-liboshmem_so_version=20:0:0
+# libopen-pal changed, added opal_random() and friends
+libopen_pal_so_version=21:0:1
+# java lib changed
+libmpi_java_so_version=20:1:0
+# liboshmem changed
+liboshmem_so_version=20:1:0

 # "Common" components install standalone libraries that are run-time
 # linked by one or more components.  So they need to be versioned as
@@ -106,5 +110,6 @@ libmca_orte_common_alps_so_version=20:0:0
 libmca_opal_common_cuda_so_version=20:0:0
 libmca_opal_common_libfabric_so_version=20:0:0
 libmca_opal_common_sm_so_version=20:0:0
-libmca_opal_common_ugni_so_version=20:0:0
+# common ugni changed
+libmca_opal_common_ugni_so_version=20:1:0
 libmca_opal_common_verbs_so_version=20:0:0
```

hppritcha:
okay I see, i'll double check and update the PR.

hppritcha:
closing this PR, @jsquyres opening another one

jsquyres:
Replaced by #1332.

PR #1370, opened 2016-09-06
Fix a typo in the COLL_SYNC macro

rhc54:
Minor correction - the macro was working in the existing code because the variable name matched in all the files. So modify it here so the macro uses the variable name in its parameter list.


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/gh-ompi-release-pr/2159/ for details.

jsquyres:
We talked about this on the call earlier today.  👍

